Issue #7, opened 2014-01-27
Why the 5.4 dependency?


mindplay-dk:
Why the dependency on PHP 5.4?

The new `Aura\Sql` is compatible with 5.3 - why the higher requirements for this component?

I see you're using traits, and maybe that's why - but the problem you're solving with traits in this case can be solved just as well using composition rather than inheritance.

Would you accept a pull-request, refactoring for PHP 5.3 compatibility (passing all unit-tests, keeping API identical to current API) or do I have to maintain my own fork?

pmjones:
@mindplay-dk -- as you say, the requirement is because of the use of traits in the query-builder portions, as well as some use of the short-array syntax. The newer v2 components are more split-up, so Aura.Sql v2 is at 5.3 (no traits), while Aura.Sql_Query is 5.4 (because of traits).

As for using composition instead of inheritance, we did try variations of composition instead of traits in Aura.Sql_Query, but it turns out to be a lot more complex than using traits for things like WHERE and LIMIT/OFFSET functionality across different query objects. Sticking with traits ended up being the easiest-to-understand solution in this case.

Obviously bug fixes are still worthy of pull requests. If you want to present changes to refactor away from traits and toward composition, that still honor all the existing tests with no changes, I'll be happy to review them.  I'm afraid I can't promise acceptance until after I see the new work.

mindplay-dk:
> Obviously bug fixes are still worthy of pull requests. If you want to present changes to refactor away from traits and toward composition, that still honor all the existing tests with no changes, I'll be happy to review them.

That's good enough for me - at least you're open to it, and there's a chance I might not have to maintain this in the future.

> I'm afraid I can't promise acceptance until after I see the new work.

Of course not :-)

I was merely probing to see if you were open to the idea at all, and apparently you're not opposed - barring I can find an elegant and backwards-compatible way to accomplish this, which I do believe I can.

I'll take a stab at it anyway. Thanks!

mindplay-dk:
> Sticking with traits ended up being the easiest-to-understand solution

Now that I'm reading the code, I have to disagree - the use of traits provides brevity, but doesn't make things easy to understand.

For example, `WhereTrait` has a dependency on `quoteNamesIn()` in it's parent class, but short of scanning the codebase and figuring out the possible parent classes, there is no way to read the code and understand where or how that dependency is satisfied.

[...]

PR #8, opened 2014-01-29
PHP 5.3 compatibility

mindplay-dk:
As per #7


src/Mysql/Delete.php
```diff
@@ -84,4 +80,33 @@ public function quick($enable = true)
...
+    /**
+     *
+     * Sets a limit count on the query.
+     *
+     * @param int $limit The number of rows to select.
+     *
+     * @return $this
+     *
+     */
+    public function limit($limit)
```
harikt, line 93: I am seeing the same code in `Delete` , `Select`, `Update` . Why not we make it as abstract class ?

src/Mysql/Delete.php
```diff
@@ -84,4 +80,33 @@ public function quick($enable = true)
...
+     * Adds a column order to the query.
+     *
+     * @param array $spec The columns and direction to order by.
+     *
+     * @return $this
+     *
+     */
+    public function orderBy(array $spec)
+    {
+        return $this->addOrderBy($spec);
```
harikt, line 110: also here, what is the advantage of using `orderBy` and `addOrderBy` . Just for no api change ?

mindplay-dk, line 110: Yes, these are all the same - they have to be, and it's not a problem, since there is only one line in every implementation, delegating to the `protected addOrderBy()` method, which contains the actual logic.

There is nothing wrong with that. People have been doing horizontal reuse this way since the dawn of classical inheritance, and it actually does have some advantages - suppose that orderBy() _wasn't_ precisely the same in one of the implementations, or it had some other side-effect: now you have an obvious place to implement those differences.

With multiple inheritance (traits) your hope is that this situation will never happen - and when it does happen (such as in the case of `getBindValues()` which had many different implementations) the code paths gets even more obscure and difficult to follow...

src/Mysql/Delete.php
mindplay-dk, line 93: There is one line of code in the `limit()` and `offset()` methods - I thought about adding another base-class, but that would add another file/class (which would be difficult to name - `AbstractSelectUpdateDeleteQuery`? yikes) and it's going to save you a total of six lines of code, all of them straight `$this->x = $y` statements with no real logic or calculations.

I also thought about adding `protected setOffset()` and `protected setLimit()` methods, and calling those - similarly to how the other traits were factored away, but again, it's still one statement (now a function call instead of an assignment) plus the return statement, in every case, so it really doesn't simplify anything, it just adds overhead.

I opted for simplicity.

harikt, line 93: Agree with the naming problem 
But repeating same code seems a bad idea though :-/.

mindplay-dk, line 93: > repeating same code seems a bad idea

DRY is not a golden rule that overrides all other programming principles - it's intended to prevent you from repeating actual code (multiple assignments, a computation, conditionals, flow control, etc.) but repeating one statement is sometimes necessary and/or better than over-abstracting in the name of DRY-ness.

Add another protected method to the base-class and you will _still_ have a single statement - now it's just a function call rather than an assignment. You're still repeating a statement - doesn't make your code more DRY.

Moreover, repeating a single assignment or function-call is extremely unlikely to cause any serious maintenance problems.

mindplay-dk, line 93: You have to weigh the advantages of clear, readable code against the value of simply keeping the number of lines down - multiple inheritance (using traits) gives you fewer lines of code, but the code demands deep analysis and reasoning to be understood; and even then, it's extremely brittle.

You should favor "simple" over "easy" - your code will thank you in the long term... adding all this boilerplate was not _easy_, but the result is very _simple_ - the code paths are clear and easy to follow. In my opinion, that's what makes good code - not how many lines or functions you had to write.

pmjones:
@mindplay-dk Dude. Duuuuuude. I was skeptical, especially of a larger AbstractQuery object, but the end result here is totally reasonable. Merging!

pmjones:
Many thanks @mindplay-dk. Bringing it back to 5.3 is a welcome move.

harikt:
oh really :+1: . Good work @mindplay-dk . Thank you.

mindplay-dk:
I'm glad you find this useful, it was a useful exercise for me as well :-)

koriym:
It's not easy job. Awesome. @mindplay-dk Congraturation !

PR #9, opened 2014-01-30
String build

mindplay-dk:
This branch factors away `AbstractQuery::$stm` avoiding accumulation of state and the previously implied order in which `build_()` methods must be called; all `build_()` methods now return strings. (first commit - passing all unit tests)

I also factored away the driver-specific `build()` methods which are no longer necessary, since the `AbstractQuery` base class is responsible for all of the actual string-building; the only exception being the MS-SQL driver, which needs to manipulate the actual statement. (second commit - passing all unit tests)

You may or may not agree with the second commit - arguably, the driver-specific `build()` methods don't add anything, but you could argue that they provide clarity, as far as which portions of a statement is actually applicable to what driver. (I understand if you prefer not to pull the second commit.)

On a related note, MS-SQL 2012 and newer might benefit from having a dedicated driver - while `LIMIT` and `OFFSET` are still not supported, something [very similar](http://dbadiaries.com/new-t-sql-features-in-sql-server-2012-offset-and-fetch) is supported since the 2012 release and forward, and you could probably override `buildLimit()` and add support for that easily. (personally I don't care or need MS-SQL support.)

### DIFF
--- a/src/AbstractQuery.php
+++ b/src/AbstractQuery.php
@@ -115,15 +115,6 @@ abstract class AbstractQuery
 
     /**
      * 
-     * The statement being built.
-     * 
-     * @var string
-     * 
-     */
-    protected $stm;
-    
-    /**
-     * 
      * The suffix to use when quoting identifier names.
      * 
      * @var string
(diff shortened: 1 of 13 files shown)


harikt:
Thanks @mindplay-dk for taking the time.

Issue #24, opened 2014-07-03
Autoquoting table names


Garfunk:
Hi,

actually tables names are auto quoted. For example, calling 

```
->from('production.some_table')
```

will result in

```
FROM "production"."some_table"
```

in the final query.

Unfortunately, sometimes (with postgreSQL at least) we may need to call a function. And function name must NOT be quoted, or it triggers an error.

```
->from('production.some_function()')
```

will result in

```
FROM "production"."some_function()"
```

but it should be

```
FROM "production".some_function()
```

I can humbly propose 2 solutions:
- autodetect a function call (by detecting the brackets at the end of the table name), but i might be not really reliable,
- adding another method/an additional parameter to stop autoquoting the table name in the `from` method.

Thanks a lot !

pmjones:
@Garfunk after trying the first suggestion, I disliked it, so I have ended up at your second suggestion. The new `Select::fromRaw()` method leaves the name unquoted. Try it out, let me know how it works for you.

Garfunk:
OK thanks a lot :)
I'll try it as soon as I can update Composer safely.

Garfunk:
I finaly got to update and I can tell you that it works like a charm, no more ugly str_replace :)
Thanks a lot !

harikt:
@Garfunk great! . Good to hear.

PR #28, opened 2014-07-08
A less magical fix for #24

pmjones:
This replaces the intention-divining logic in the Quoter with a more straightforward `Select::fromRaw()` method. Using this method means the name will not be quoted.


tests/src/QuoterTest.php
```diff
@@ -47,25 +47,4 @@ public function testQuoteNamesIn()
         $expect = "*, *.*, `foo`.`bar`, CONCAT('foo.bar', \"baz.dib\") AS `zim`";
         $this->assertSame($expect, $actual);
     }
-
```
harikt, line 50: Why the test need to be removed?

pmjones, line 50: Because the quoter no longer applies that kind of quoting (i.e., to functions).

PR #33, opened 2014-07-26
Alternative key-cols-by-alias logic

pmjones:
@mbrevda there were some issues with your PR:
- did not address "col alias" (i.e., space as AS, which is allowed)
- it looked like it might overwrite sequential keys on subsequent calls to cols() but I did not test it
- the test itself looked backwards to me; it should be "col AS alias"

I have added the "key $cols by the alias when present" functionality in an alternative fashion in this PR. There is one "gotcha" and it is that later uses of the same alias name will overwrite earlier uses; I can see where this would be unexpected behavior at the PHP level (previously it would have resulted in an SQL error).

### DIFF
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -233,7 +233,11 @@ class Select extends AbstractQuery implements SelectInterface
      */
     protected function addCol($key, $val)
     {
-        $this->cols[] = array($key, $val);
+        if (is_int($key)) {
+            $this->cols[] = $val;
+        } else {
+            $this->cols[$val] = $key;
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mbrevda:
1. Correct, I didn't know that was valid
2. It will. Such is the nature of arrays. Due to Aura's lightweight approach, I think that is ok

LGTM. Lets get merged and Ill start working on the next part (i.e. manipulation)

src/Common/Select.php
```diff
@@ -233,10 +233,27 @@ public function cols(array $cols)
      */
     protected function addCol($key, $val)
     {
-        if (is_int($key)) {
-            $this->cols[] = $this->quoter->quoteNamesIn($val);
```
mbrevda, line 237: it would seem that numeric aliases are also valid. Why not pass in the $key?

pmjones, line 237: Have to allow for $select->cols(array('foo', 'bar', 'baz')) without it converting to "foo AS 1, bar AS 2, baz AS 3".

mbrevda, line 237: got it

jakeasmith:
FYI this change caused several of our tests to fail. We had several queries that like this: `->cols(['distinct col1', 'col2])` that suddenly turned into `distinct as col1`. It was an easy fix after I noticed this distinct method, but it was a little weird to track down. Hope this helps someone.

Old code: `$select->cols(['distinct col1', 'col2])`

New code: `$select->distinct()->cols(['col1', 'col2])`

harikt:
Thank you @jakeasmith for reporting, and showing the fixed stuff.

PR #39, opened 2014-10-29
MySQL: INSERT ... ON DUPLICATE KEY UPDATE

kornrunner:
Is there a plan for supporting `INSERT INTO ... ON DUPLICATE KEY UPDATE` variaton of `INSERT` for MySQL? I'd attempt to work it out and send a pull request if there's any interest.

### DIFF
--- a/src/Common/Insert.php
+++ b/src/Common/Insert.php
@@ -123,7 +123,7 @@ class Insert extends AbstractDmlQuery implements InsertInterface
      *
      * @param string $col The column name.
      *
-     * @param mixed  $val Optional: a value to bind to the placeholder.
+     * @param mixed,...  $val Optional: a value to bind to the placeholder.
      *
      * @return self
      *
--- a/src/Mysql/Insert.php
+++ b/src/Mysql/Insert.php
@@ -201,7 +201,7 @@ class Insert extends Common\Insert
 
         $values = array();
         foreach ($this->col_on_update_values as $key => $row) {
-            $values[] = $this->indent([$key . ' = ' . $row]);
+            $values[] = $this->indent(array($key . ' = ' . $row));
         }
 
         return ' ON DUPLICATE KEY UPDATE'
--- a/src/Mysql/Insert.php
+++ b/src/Mysql/Insert.php
@@ -199,6 +199,7 @@ class Insert extends Common\Insert
             return ''; // not applicable
         }
 
+        $values = array();
         foreach ($this->col_on_update_values as $key => $row) {
             $values[] = $this->indent([$key . ' = ' . $row]);
         }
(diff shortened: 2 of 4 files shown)


kornrunner:
I've added the functionality, implemented three new methods:
`colOnUpdate`
`colsOnUpdate` and
`setOnUpdate` - to match existing functionality.

I've also had to hold column values in different array for that part of query, and changed bind names for those columns - to avoid mixing with original insert values... Not the happiest solution I think, do you have some other suggestion?

Also, I'm not really satisfied with method names and with the tests, but it's something... Please let me know what do you think. Thanks!

src/Mysql/Insert.php
```diff
@@ -80,4 +90,121 @@ public function delayed($enable = true)
...
+
+    /**
+     *
+     * Sets one column value placeholder in ON DUPLICATE KEY UPDATE section;
+     * if an optional second parameter is passed, that value is bound to the
+     * placeholder.
+     *
+     * @param string $col The column name.
+     *
+     * @param mixed $val Optional: a value to bind to the placeholder.
```
staabm: Not existent

kornrunner: This was supposed to be equivalent of `col` method from https://github.com/auraphp/Aura.SqlQuery/blob/develop-2/src/Common/Insert.php#L131 and code is almost the same as that method - I've left the same phpdoc as well (it's using `func_get_args`)

staabm: I see. per phpdocumentor this should read `mixed,...` instead of `mixed`

src/Common/Insert.php
```diff
@@ -123,7 +123,7 @@ public function getLastInsertIdName($col)
      *
      * @param string $col The column name.
      *
-     * @param mixed  $val Optional: a value to bind to the placeholder.
+     * @param mixed,...  $val Optional: a value to bind to the placeholder.
```
harikt, line 126: oh I wasn't aware we need `,...` .

harikt:
Thank you @kornrunner for your work. @pmjones will look and merge. I am ok with this for I don't have a use case currently.

Question I have in mind is 
1. Hope it will not be having any problem passing the query to PDO. 

May be we need to document this if @pmjones is satisfied before merge. So you can wait to hear him before your work :+1:  .

pmjones:
@kornrunner I changed the method names to `onDuplicateKeyUpdate()`, `onDuplicateKeyUpdateCol()`, and `onDuplicateKeyUpdateCols()` (for set, col, and cols, respectively), and used suffixes instead of prefixes on the modified bind placeholder names. Thanks!

kornrunner:
@pmjones Thank you very much!

Issue #40, opened 2014-11-26
Limit, offset and paging


Grummfy:
Hi,
I find bad that if we directly set offset and limit to a select query, that getPaging() doesn't give an updated value.

By the way, I think it could be really intresting to have a getLimit() and a getOffset() method on Select . Why ? Because it's easier to extends, easier to test, ...

harikt:
I believe you should make use of 

``` php
<?php
$query->setPaging(10);
$query->page(2);
```

which should set the offset and limit for you.

harikt:
Here is what the code does https://github.com/auraphp/Aura.SqlQuery/blob/2.0.0/src/Common/Select.php#L474-L489 . Even if you don't set the `setPaging` by default it takes 10.

Grummfy:
Yes, but without using setPaging, you can't have a getPaging relevant. So for me it's a bit confusing.

harikt:
@Grummfy https://github.com/auraphp/Aura.SqlQuery/blob/2.0.0/src/Common/Select.php#L104 :-)

Grummfy:
OKi, I just understand my confusion. getPaging() is the number of row per page (it's writen in the comment ...) and mixed it with a getPage() or whatever ... and my code. Well ok, my bad.

But for me, it should have two more accessors :
- getLimit()
- getOffset()

harikt:
If you feel so, you should consider sending a PR and wait for Paul's thoughts or pressure him to merge ;) .

PR #41, opened 2014-11-27
resolve issue #40 aka adding getter on limit and offset

Grummfy:
Here is the #40 PR

### DIFF
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -561,6 +561,16 @@ class Select extends AbstractQuery implements SelectInterface
         return $this;
     }
 
+	public function getLimit()
+	{
+		return $this->limit;
+	}
+
+	public function getOffset()
+	{
+		return $this->offset;
+	}
+
     /**
      *
      * Clears the current select properties; generally used after adding a
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -561,15 +561,15 @@ class Select extends AbstractQuery implements SelectInterface
         return $this;
     }
 
-	public function getLimit()
-	{
-		return $this->limit;
-	}
+    public function getLimit()
+    {
+        return $this->limit;
+    }
 
-	public function getOffset()
-	{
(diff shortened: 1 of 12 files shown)


src/Common/Select.php
```diff
@@ -561,6 +561,16 @@ public function unionAll()
         return $this;
     }
 
+	public function getLimit()
+	{
+		return $this->limit;
```
harikt: seems tabs :) .

harikt:
Thank you @Grummfy . You have used 1 tab = 8 character. Please change to 4 space.

Paul will look into this.

Grummfy:
1 tabs = how many characters you want ... it's a configuration of your editor ...
but I change all my tabs to 4 space ... it will be easier if you have a checkstyle file to run phpcsfix...

harikt:
1 tab = 4 charactes .

Thanks!

_Hari K T_

You can ring me : +91 9388 75 8821

http://harikt.com , https://github.com/harikt ,
http://www.linkedin.com/in/harikt , http://www.xing.com/profile/Hari_KT

Skype  : kthari85
Twitter : harikt

staabm:
this is a BC break, should it require a new major version?

Grummfy:
@staabm  nop. jsut add some getters

staabm:
@Grummfy you changed existing interfaces, therefore classes which implement them outside of the lib will get a fatal error after this PR lands

harikt:
@staabm good call :) .

pmjones:
So, is this a BC break or not?

harikt:
@pmjones it seems to be a BC break :-/ . If someone have implemented `LimitInterface.php` and `LimitOffsetInterface.php` they should implement it in those libraries. May be ask this guy https://github.com/tomzx/php-semver-checker

pmjones:
Merged manually, removed the Interface changes that would have been BC breaks.

PR #42, opened 2015-01-11
convert ? placeholders in where()/having() to :_#_ named placeholders

pmjones:
This is because PDO is touchy about numbering on `?` placeholders. If we have bound values `[:foo, :bar, ?, :baz]`, the ? placeholder is not number 1, it is number 3. As such it is nearly impossible to keep track of the numbering when values are bound out-of-order.

This PR modifies the code to do a braindead check on the where/having condition string to see if it has `?` placeholders, and replace them with named `:_#_` placeholders, where `#` is the current count of the `$bind_values` array. It eliminates the problem of keeping proper numbers on `?` placeholders.

@mbrevda and @mindplay-dk especially take note, and please review at your convenience. The vast majority of changes is in the tests, from `?` to `:_1_` etc.

### DIFF
--- a/src/AbstractDmlQuery.php
+++ b/src/AbstractDmlQuery.php
@@ -40,25 +40,6 @@ abstract class AbstractDmlQuery extends AbstractQuery
 
     /**
      *
-     * Gets the values to bind to placeholders, including any 'where' values
-     * (needed for INSERT and UPDATE).
-     *
-     * @return array
-     *
-     */
-    public function getBindValues()
-    {
-        $bind_values = parent::getBindValues();
-        $i = 1;
-        foreach ($this->bind_where as $val) {
-            $bind_values[$i] = $val;
-            $i ++;
-        }
-        return $bind_values;
-    }
-
-    /**
-     *
      * Sets one column value placeholder; if an optional second parameter is
      * passed, that value is bound to the placeholder.
      *
--- a/src/AbstractQuery.php
+++ b/src/AbstractQuery.php
@@ -42,15 +42,6 @@ abstract class AbstractQuery
 
     /**
      *
-     * Bind these values to the WHERE conditions.
-     *
-     * @var array
-     *
-     */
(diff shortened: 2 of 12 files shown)


pmjones:
@mbrevda I mistyped your Github name earlier, so you probably did not get a notification. Any thoughts on this PR?

mbrevda:
LGTM. I haven't tested it against my stuff, but it doesn't look it will break anything. I wish you took an official stance on php5.3 though (i.e. lot's of Aura's libs claim to support php5.3, yet don't pass the tests)

Would it be clearer to _name_ the placeholders? i.e. `:placeholder1, :placeholder2`, etc

pmjones:
> Would it be clearer to name the placeholders? i.e. :placeholder1, :placeholder2, etc

No matter sane name I pick, some other sane person will be using it. :-/

> lot's of Aura's libs claim to support php5.3, yet don't pass the tests

Which ones? If they don't, I need to know.

mbrevda:
I was referring to this, which seems to be fixed , in addition to your
tests here which I though might have intentional:

https://travis-ci.org/auraphp/Aura.Filter/builds/46733972

mindplay-dk:
Mixing named and nameless placeholders is a sad, terrible, bad bad idea. Personally I wouldn't even attempt to define a behavior for that, I would just throw an exception. Just my two cents.

pmjones:
@mindplay-dk:

> sad, terrible, bad bad idea

IOW "It's bad, and you should feel bad." ;-)

The behavior of "bind a value to a ? in a where()/having() condition" is already established in a stable release, so I'd need to release a v3 to throw an exception instead of doing _something_ to retain that behavior.

pmjones:
@mbrevda: no, if a library says "5.3 compatible" but is not, it's either a failure of coding, or a failure of documenting. In that case it looks to have been a failure of coding.

mbrevda:
OK, got it

PR #43, opened 2015-01-17
Fix quote issue with single character table alias

aaronwebb:
When a name contained a table alias of one character it wasn't getting quoted.


pmjones:
Good catch!

PR #44, opened 2015-02-03
Add ability to SELECT Max (field)

MarcusFulbright:
I've been using Aura.SqlQuery at work for the past two weeks. We have to do some `Select Max (field) From <table>` queries. It'd be nice to use this library for those queries as well.

### DIFF
--- a/src/Common/SelectInterface.php
+++ b/src/Common/SelectInterface.php
@@ -232,4 +232,15 @@ interface SelectInterface extends QueryInterface, WhereInterface, OrderByInterfa
      *
      */
     public function unionAll();
+
+    /**
+     *
+     * Will select max value from a specified column
+     *
+     * @param string $col name of the Column
+     *
+     * @return self
+     *
+     */
+    public function max($col);
 }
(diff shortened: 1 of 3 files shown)


src/Common/SelectInterface.php
```diff
@@ -232,4 +232,15 @@ public function union();
...
+    /**
+     *
+     * Will select max value from a specified column
+     *
+     * @param string $col name of the Column
+     *
+     * @return self
+     *
+     */
+    public function max($col);
```
staabm, line 245: This breaks BC

MarcusFulbright, line 245: The functionality was implemented in Common/Select. All select classes inherit from that correct? Where is the BC break?

pmjones:
/me thinks

Have we tried `$select->from('table')->cols(['MAX(foo) AS max_foo'])` ?

MarcusFulbright:
@pmjones that passes the same test in this pull request (just requires a slight whitespace change). Somehow I missed the `count()` example in the docs too. Totally my bad.

Now that I know functions can be in the `$cols` array, implementing `$select->max('col')` feels like a slippery slope that leads to implementing all the functions. 

Good thing this pull request only took a few minuets to code.

PR #53, opened 2015-03-06
Fixed tests for windows

ksimka:
In tests you're trying to compare strings built with PHP_EOL with strings written in test files with fixed `\n` line endings. That leads to tests failures on windows where PHP_EOL equals to `\r\n` (and mac I suppose, where PHP_EOL equals to `\r`).

So it's better to remove all the line endings in query strings.


pmjones:
I took the liberty of modifying this to use `PHP_EOL` instead of `' '`. Otherwise the PHPUnit output comparisons will be really wide.

ksimka:
Ok, thanks.

But as for me, checking too strict is not necessary.
For SQL server both

```
SELECT *
FROM table
```

and 

```
SELECT * FROM table
```

are equal.

So if you will decide to create some formatter for sql builder (for ability to get both formatted and short versions of query) your tests will fail or you will need to create exactly the same formatter as now.

Issue #16, opened 2014-04-01
Insert multiple rows with a single query


jakeasmith:
Is there a way to build a query that inserts multiple value sets? 

`INSERT INTO tbl_name (a,b,c) VALUES(1,2,3),(4,5,6),(7,8,9);`

If not, do you think it makes sense to add that functionality?

harikt:
This is a nice functionality to add. Currently we don't have this support.

harikt:
I also love to have multiquery execution. But we already discussed this in Aura.Sql https://github.com/auraphp/Aura.Sql/issues/22

pmjones:
I can see where this would be useful, but I'm afraid I have other things that have to take priority.  However, if you want to create a new MultipleInsert (or InsertMany or InsertMulti) class that exhibits this behavior and send a PR for it, I'd be happy to review.

arcturial:
I am working on a prototype for doing bulk inserts.

https://github.com/arcturial/Aura.SqlQuery/tree/feature-bulk-insert

Once the idea is solid I will do some cleanup before doing a PR. Have a look when you have a moment to see if this is what you had in mind.

The way you would use it is as follows:

``` php
$query->newBulkInsert()
           ->cols(array('col1', 'col2', 'col3'))
           ->bindValues(array(
                    array(
                            'col1' => 'row1a',
                            'col2' => 'row1b',
                            'col3' => 'row1c',
                    ),
                    array(
                            'col1' => 'row2a',
                            'col2' => 'row2b',
                            'col3' => 'row2c'
                    )
             ));
```

harikt:
@arcturial it looks good to me. Appreciate your good work on it.

arcturial:
I moved the code to a new branch and changed it so that it now uses the same interface as the Insert() queries. 

https://github.com/arcturial/Aura.SqlQuery/tree/feature-bulk-insert

The behaviour of some of the methods had to change slightly (like `cols()` not supporting associative arrays) since we are dealing with bulk data and not single values, but I don't think this should cause usability problems. I am not really sure why someone would want to use `cols()` with an associative array when doing bulk inserts anyway.

arcturial:
I updated the README in that branch with more detailed usage instructions.

harikt:
@arcturial send a PR so it will be easy to get reviewed.

Thanks!

arcturial:
https://github.com/auraphp/Aura.SqlQuery/pull/56, request created

[...]

PR #56, opened 2015-03-19
Implement bulk insert capability. issue #16

arcturial:
BulkInsert handler for Common, MySQL, Postgress, SQLserv, SQLite.


harikt:
:+1:  good work dude .

src/Common/BulkInsert.php
```diff
@@ -0,0 +1,154 @@
...
+     * @var array
+     */
+    protected $cols = array();
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function addCols(array $cols)
+    {
+        foreach ($cols as $key => $val) {
```
stof: `$key` is unused

arcturial:
@stof, removed `$key`

pmjones:
Good on you, @arcturial, for doing the work. I will review and respond over the next few days.

pmjones:
Hi @arcturial -- I began reviewing your PR this morning. It looked fine, but needed some touching up. While I was doing the touching-up work I had a revelation that would allow the BulkInsert to be integrated into the regular Insert. It's a major modification from your PR, but I would not have thought of it if I had not seen your work first. Please take a look at #57 and see if it suits your needs.

Usage is as in the added test in that PR. The order-of-columns is taken from the first added row. You call `addRow()` to add another row, which collects and then resets the insert values. It used named placeholders instead of sequential ones, and renames them for the row number.

PR #57, opened 2015-03-21
add bulk insert logic to Insert, thanks to @arcturial for his initial work

pmjones:


arcturial:
Hey, looks good. I initially thought of integrating it with `Insert` itself, but was considered about backwards compatibility. I do have a few suggestions though.

I would suggest you also add something like a helper function to create a query based on an associative array. Right now you would have to do something like:

``` php
$data = []; // array of N rows
$first = array_splice($data);

$insert = $factory->newInsert()
      ->cols($first);

foreach ($data as $row) {
    $insert->addRow();
    $insert->cols($row);
}

```

If you include the appropriate error checking (does the data array contain a first element etc) it can get quite cumbersome if you have to duplicate that logic everytime. Some like the following might be simpler to use:

``` php

$data = []; // 50 elements
$row = []; // single row

$query = $factory->newInsert()
    ->addRows($data) // returns the object pointing to the row after the last one that was added
    ->addRow($row) // returns the object at the next row

```

It doesn't have to change the functionality of `addRow`, but something that abstracts and eliminates the need for looping from the users side would be nice.

Then another thing, not sure if you are interested in supporting this...but some backwards compatibility for SQLite bulk inserts would be nice. Only the newer versions of SQLite supports the same bulk insert syntax as the rest of the SQL's.

pmjones:
> a helper function to create a query based on an associative array. 

Good call. Let me see what I can do there.

> some backwards compatibility for SQLite bulk inserts would be nice

That one I'm not so sure about. Maybe something specific in the Sqlite Insert object. It'd take some more research.

UPDATE: This looks like the SQLite stuff: http://stackoverflow.com/questions/1609637/is-it-possible-to-insert-multiple-rows-at-a-time-in-an-sqlite-database/5009740#5009740

Really, only 3.7.11 and later supported the "standard" bulk insert? Man.

pmjones:
All right, @arcturial , check this out: `addRow()` now optionally takes an array of `$cols` values, and the new method `addRows()` takes an array of `$rows`. The returns are a little different from what you detailed; let me know if it suits you.

arcturial:
There is a workaround in the older versions of SQLite that allows you to do bulk inserts using unions, but that is something I can have a look at adding later. I would suggest having the `set()` method "override" values when in a bulk inserting context.

``` php
$query->addRows(array(
    array('col1' => 'val1', 'col2' => 'val2'),
    array('col1' => 'val2', 'col2' => 'val2')
));

$query->set('col2', 'NOW()');
```

This will result in all the values for `col2` to be replaced by NOW(). If col2 wasn't specified in the associative array, it will be added. I am not sure why you would have a situation where you would want to bind a value to one row and then use a fixed value in the next (unless I am missing something). Doing it this way allows for a more robust and shouldn't cause unexpected problems depending on the order you execute them. Besides that, looks good.

pmjones:
> There is a workaround in the older versions of SQLite that allows you to do bulk inserts using unions, but that is something I can have a look at adding later.

Yeah, I saw that in the StackOverflow link. I agree it can wait until later. (The first problem I can imagine is, how to tell the Sqlite Insert object which variation to use.)

> I would suggest having the set() method "override" values when in a bulk inserting context. ... I am not sure why you would have a situation where you would want to bind a value to one row and then use a fixed value in the next

One never knows, now, do one?  Every time I try to do clever helpful things like that, it comes back to bite me. :-/  Having said that, it may be that an additional method `setInAllRows()`, or something like that, may be a useful addition. Do you feel it is critical functionality for the next feature-release? (With the spate of changes this week I think a new release is warranted.)

arcturial:
You are absolutely right, there's always that one use case that screws you. It's not a high priority feature (you can always just loop if you need it). However, having the ability would be nice...maybe a third optional boolean parameter like `set($col, $value, $allRows = false)` could work too.

pmjones:
/me nods

We can discuss details on bulk-set later, then.

Here's my next problem. I want to make sure you get automatic credit via Github for having submitted the PR that led to this one, because this one simply would not exist without you having done the prior work. So the plan is to merge yours locally, then merge _this_ one on top of it, resolving conflicts along the way. Do you feel all right with that?

arcturial:
Sounds good to me. When you are done I will create a fresh PR for the SQLite workaround.

pmjones:
Good deal! Starting on the messy merge now.

PR #59, opened 2015-03-24
Add rudimentary array support for Postgres

gzumba:
This is some very basic support, ie. no escaping and no support for decoding returned arrays, as it is quite hard to do automatically.

I don't like how the bindValue implementation is duplicated, but there is no nice place to put it in and I did not want to create a new helper or base class for this purpose.

### DIFF
--- a/src/Pgsql/Insert.php
+++ b/src/Pgsql/Insert.php
@@ -54,4 +54,21 @@ class Insert extends Common\Insert implements Common\ReturningInterface
     {
         return $this->addReturning($cols);
     }
+
+    /**
+     * Bind a value, supporting simple array values. No escaping done, whatsoever.
+     *
+     * @param string $name
+     * @param mixed $value
+     * @return self
+     */
+    public function bindValue($name, $value)
+    {
+        if (is_array($value)) {
+            $value = '{' . implode(",", $value) . '}';
+        }
+
+        return parent::bindValue($name, $value);
+    }
+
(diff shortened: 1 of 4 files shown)


pmjones:
This strikes me as the kind of thing that would work better as a pre-filter or a passthru-filter in a Gateway, rather than something that goes in a query object. Do you get what I mean there?

gzumba:
Basically that would mean I need an additional DB specific layer between my Data mapper classes and Aura. Which would not be too bad and would solve the code duplication issue and at the same time the escaping/decoding issues as you would have all the field type information available.

I proposed this as a convenient shortcut for the basic case, which would make the code very nice:

```
$insert->into(self::TABLE_NAME)
->cols($product->getData())
->returning([Product::ID_FIELD])
;
```

So, this would basically be the same kind of help as ExtendedPDO is providing for array values (IN (?)).

Anyway, no big deal, I will code around this.

pmjones:
I appreciate your considerate reply; thank you.

> an additional DB specific layer between my Data mapper classes and Aura.

I applaud you for using a Data Mapper.  FWIW, I have started using a Gateway class "beneath" my Mappers, specifically for transforming values into database-specific structures. Those transformed values are then passed to Aura.SqlQuery. I started doing that mostly because of Postgres being touchy about booleans, but I find it applies to all sorts of things now that the Gateway is in place.

PR #68, opened 2015-05-25
Add getStatement method to QueryInterface

thinkspill:
Since all methods now have getStatement since 2.3, add to QueryInterface to help with IDE autocompletion.

### DIFF
--- a/src/QueryInterface.php
+++ b/src/QueryInterface.php
@@ -76,4 +76,13 @@ interface QueryInterface
      *
      */
     public function getBindValues();
+    
+    /**
+     *
+     * Returns this query object as an SQL statement string.
+     *
+     * @return string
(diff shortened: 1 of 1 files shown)


harikt:
You cannot change the interface which is BC break. Wait for 3.x for it is almost there.

thinkspill:
Ah, thanks very much. Waiting patiently. ;)

harikt:
:) .

pmjones:
As @harikt notes, we can't change an Interface without breaking BC. However, adding `@method` hints to the DocBlock might cause the IDE to "see" the method. You want to try that out @thinkspill ? Cf. http://phpdoc.org/docs/latest/references/phpdoc/tags/method.html

Issue #84, opened 2016-02-23
Allow reset for some query parts


golgote:
What about:
public function orderBy(array $spec, $replace = false)
instead of : public function orderBy(array $spec)

This can be added to other clauses as well.
I personally need this because my query is built by different objects and each adds some clauses that might later need to be replaced by others.

harikt:
How about send a PR and in case @pmjones finds the time he can merge if that is ok :-) .

golgote:
OK I will do that ASAP. I was just checking if you were open to this change.
Thanks.

yespire:
Which syntax is better:
1. 
single reset() method handle all cases
query->reset()   // -- all, paramter is mixed optional
query->reset('order_by') // -- single factor
query->reset(['order_by', 'cols']  // -- multi-factors

2.
a set of specialized reset methods
query->reset() // -- all, no parameter
query->resetCols() // -- no parameter
query->resetOrderBy()

3.
support both 1 + 2

harikt:
PR #91  .

pmjones:
There are now 2 PRs in for this issue: #91 and #94. I will pick one of them, or implement the functionality myself. Thanks for your patience!

pmjones:
@golgote @yespire @gauthier @harikt --

I am considering a series of `without*()` methods that return a clone of the current query. E.g.:

``` php
$newQuery = $oldQuery->withoutWhere()->withoutCols()->withoutOrderBy();
```

That way the existing query continues to exist, and the new one can be independently modified.

Thoughts?

pmjones:
No, on further consideration, we already have some `reset*()` methods that can be exposed without BC break. I'll expand on those instead. If the user wants a clone, they can clone.

golgote:
I like this idea, but sometimes one might just need to modify the original query, not create a new clone.

harikt:
@pmjones FYI in case if you missed the discussion on https://github.com/auraphp/Aura.SqlQuery/pull/91#issuecomment-217611396 .

pmjones:
Addressed by #95 et al. Let me know if those do not server your purposes.

PR #91, opened 2016-05-07
reset methods

yespire:


src/Common/Select.php
```diff
@@ -751,6 +751,34 @@ protected function reset()
     }
 
     /**
+     * if user needs to reset
+     */
+    public function resetCols()
+    {
+        $this->cols = [];
```
harikt: php 5.3 , don't allow short array. So the tests is failing.

harikt:
Do we need to return `$this` for method chaining ? Looking at other methods.

Additionally we need to add a few tests.

yespire:
this package seems assuming no chaining in most method

staabm:
Whats the use-case? Resetting objects is an anti pattern..

yespire:
actually there don't seem to be a consistent return vs non-return $this

some methods return $this --- ie: setPaging()
some methods doesn't return $this --- ie: reset()

yespire:
@staabm 

could you please elaborate "unsetting object is an anti-pattern" ?

#84 explained a use case

but here is another one:

the packages calcFoundRows() doesn't have a paired up method and property to capture "SELECT FOUND_ROWS()", even if it is added, some people may prefer to not use that and use SELECT COUNT(*)
//
// now imagine we need to get a list of records, then get the potential record count, the would be like:
//
$recordset = $query->fetchRecordSet();

$query->resetCols()
$query->resetOrderBy()
$query->resetLimit()
$query->resetOffset()
$query->cols([ "count(*)" ])
$listing_total = $query->fetchValue();

staabm:
You should add the counterpart instead of relying on impl. Details and partly reset the object .

Noone can guarantee that the query objects will behave the way you need it across versions

yespire:
@staabm 

That example is not implying don't add calcFoundRows()'s getter

IF there is no edit/unset/reset methods, the query building process become "add-only"
Can you explain why edit / unset / reset part of query object is bad ?

staabm:
Objects can depend on several properties and additional properties could be added in the future. In case you rely on resettig certain properties your logic can/will start failing when the object your are resetting gets new properties in the future (which can happen even in minor releases)

yespire:
@staabm

What you said is true "in general", However I don't see that applies here.

1.
Cols/OrderBy/Limit/Offset of Query/Select should be pretty stable, which probably won't change in near future. Maybe you are referring to something else ?

2.
FOUND_ROWS() has its own issues, implementing counterpart methods for Mysql/Select :: calcFoundRows() may not be helpful, there is an discussion regarding this:
https://github.com/atlasphp/Atlas.Orm/issues/5

3.
Only parent query builder object is depending on these "Cols, OrderBy, Where, etc". when we use query builder, we normally go through 2 steps: constructing phase, which business logics are checked to add / remove / modify the query builder states/properties, when all logics are checked, it builds a statement. The query builder only impact other objects via final statement. Ability to reset / modify Cols, OrderBy, Where is necessary feature for a query builder.

pavarnos:
Could we work around the issues @staabm raises by adding a selectTotalItemCount() method which internally resets the correct parts of the query, adds a COUNT(*) or SQL_CALC_FOUND_ROWS? This would hide the implementation details he is concerned about and make pagination easy...

harikt:
@pavarnos I like the idea.

gauthier:
This is huge! I just pushed a very similar PR, that I had in mind for a while, but I did not check someone already did!

Sorry for the duplicate PR :(

Issue #69, opened 2015-05-28
Remove restriction requiring FROM specification before JOIN


elazar:
Presently, when building a `SELECT` statement, `from()` [must be called](https://github.com/auraphp/Aura.SqlQuery/blob/2dff47fd765814705813364f3af6c92c679d2214/src/Common/Select.php#L401-403) before `join()`. My understanding is that this is more for semantic (i.e. `JOIN` essentially being a sub-clause of `FROM`) than logical / logistical reasons. There are [use cases](https://twitter.com/elazar/status/603957835957837824) that would be supported by not having this restriction.

While I may attempt implementing this myself at some point, I'm uncertain of how soon that will be. As such, I'm filing this issue in case anyone else runs into the same situation and/or wants to try to implement it themselves.

pmjones:
Well, it's been a year. @elazar do you feel this is still something that's needed? Or is it a doc fix, where we say "you must call from() before join()" ?

elazar:
I'd call it a "nice to have." If a code change isn't made, a doc change probably should be.

pmjones:
@elazar Please review PR #96 , which is an attempt at fixing this issue.

pmjones:
Fixed by #96; please open a new issue if it doesn't work the way it should. Thanks!

PR #96, opened 2016-05-19
Allow join*() before from*()

pmjones:
@elazar @metalfox6383:

This PR allows join() before from(). Joins-before-from are added to the first from(). If no from is ever added, the joins will never be built into the statement. Please try it out and see if it works as you expect.

### DIFF
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -65,6 +65,8 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
      */
     protected $from_key = -1;
 
+    protected $join = array();
+
     /**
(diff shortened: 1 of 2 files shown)


elazar:
@pmjones Damn, I wish I'd captured more context in #69. From what I can remember offhand, though, it looks like this would have worked for what I was doing. I'd probably lean more on @metalfox6383 since their use case appears more recent / immediate.

pmjones:
Command decision, then: MERGE!

compwright:
@pmjones any chance you could get this into a 2.6 release? I seem to be running into a related problem.

Issue #74, opened 2015-06-15
Fatal error: Can't inherit abstract function Aura\SqlQuery\Common\SubselectInterface::getBindValues() (previously declared abstract in Aura\SqlQuery\QueryInterface) ...


arvidj:
I get this error in PHP 5.3.2. 

<code>
Fatal error: Can't inherit abstract function Aura\SqlQuery\Common\SubselectInterface::getBindValues()(previously declared abstract in Aura\SqlQuery\QueryInterface) in vendor/aura/sqlquery/src/Common/Select.php on line 22
</code>

Seems to working fine in 5.4.36. Downgrading to SqlQuery 2.4.2 solves the problem.

pmjones:
@arvidj This sounds like it might a language-level issue. Can you tell what the earliest version of PHP 5.3 is where the error does not occur?

sserbin:
This was fixed in 5.3.9 (https://bugs.php.net/bug.php?id=43200 , http://php.net/ChangeLog-5.php#5.3.9) - probably makes sense to bump composer's minimal version requirement.

harikt:
@sserbin thank you for pointing to it. I have made a https://github.com/auraphp/Aura.SqlQuery/pull/99 . If @pmjones agrees we can merge . In the mean time I also want to see if we need to make any travis change for it.

pmjones:
(I seem to recall having a discussion like this elsewhere.)

Changing a minimum version number in `composer.json` strikes me as a BC break. If you're still on something earlier than 5.3.9, then doing a `composer update` after we update the minimum PHP version will cause a failure to install, even though you haven't changed anything about your own system.

For something as old as 5.3.x, do we actually care about this?  I am beginning to think not, but I'd like to hear your opinions.

harikt:
I don't care much about older versions for they have reached EOL. But I do understand the pain of those developers who work on it. So if the library is supporting the older versions then bugs should be fixed.

For the newer versions I am for PHP 7+ ;) .

pmjones:
Fixed by @harikt in #99 . Breaking my own rules on BC breaks, but I think this one very limited case is, if not exactly "acceptable", then at least "understandable."

PR #99, opened 2016-07-18
Due to bug https://bugs.php.net/bug.php?id=43200

harikt:
Which is fixed in 5.3.9 http://php.net/ChangeLog-5.php#5.3.9 . Thank you @sserbin for pointing the same over  https://github.com/auraphp/Aura.SqlQuery/issues/74#issuecomment-233289858


pmjones:
Breaking my own rules. :-/

harikt:
Rules are to be broken not to be tightly hold :+1:  .

PR #109, opened 2016-12-30
insert->col() and update->col() use variable argument lists 

pavarnos:
so IDEs will not show an error if the $value parameter is used.


pavarnos:
oops. PR against 2.x instead of 3.x branch. Sorry!

PR #110, opened 2016-12-30
fix phpDoc and broken return type in Common/Select.php

pavarnos:


src/Common/Select.php
```diff
@@ -248,8 +248,6 @@ protected function addCol($key, $val)
      * @param string $spec The column specification: "col alias",
      * "col AS alias", or something else entirely.
      *
-     * @return null
```
harikt, line 251: So what is the return type we can define ? void ?

src/Common/Select.php
```diff
@@ -273,7 +271,7 @@ protected function addColWithAlias($spec)
      *
      * @param string $alias The column to remove
      *
-     * @return null
+     * @return boolean
```
harikt: Inconsistency in return type : https://github.com/auraphp/Aura.SqlQuery/pull/110/files#diff-57cf575fdd7a21c19685563c9fa2a1c0L300

src/Common/Select.php
pavarnos: I was fixing a phpStorm lint warning that said "missing return statement". Didn't notice all the other similar occurrences in the file and throughout the code. I guess for consistency, it should `@return null` or `@return void` everywhere. What is your preference?

harikt: return null is all good for now I believe.

Thank you

src/Common/Select.php
harikt: This is not yet fixed. Why not use `bool` ?

pavarnos: bool and boolean are the same, (http://php.net/manual/en/language.types.boolean.php). Changed it to bool. The old return type of null was incorrect: every return value is either `true` or `false`.

harikt:
Thank you.

Issue #100, opened 2016-08-11
REPLACE sql command


pavarnos:
Hi

Would you accept a pull request that allowed the hard-coded "INSERT" statement in */Insert.php to be swapped for "REPLACE" when we need to do a REPLACE INTO query? (supported by sqlite, mysql)

Would probably implement it like this:

``` php
private $insertCommand = 'INSERT';

public function useReplace() 
{
    $this->insertCommand = 'REPLACE';
}

protected function build()
{
        return $this->insertCommand
            . $this->buildFlags()
            . $this->buildInto()
            . $this->buildValuesForInsert()
            . $this->buildValuesForUpdateOnDuplicateKey()
            . $this->buildReturning();
}
```

I prefer this approach over onDuplicateKeyUpdate() because it allows my tests to run unchanged on sqlite while using mysql on the live server. On Duplicate Key update is mysql only?

pmjones:
I'd be happy to look at a PR for this, if you're willing to submit one! I'd figure it should use a flag on the `useReplace()` method so it could be switched back and forth. E.g., `$insert->useReplace(false)` would put it back to a regular insert.

pmjones:
Also, it should probably be on the db-specific INSERT classes, not the common one.

PR #111, opened 2016-12-30
insert->col() and update->col() use variable argument lists

pavarnos:
so IDEs will not show an error if the $value parameter is used.


src/Common/Update.php
```diff
@@ -119,12 +119,13 @@ public function orWhere($cond)
      *
      * @param string $col The column name.
      *
-     * @return $this
+     * @param array  $value
```
harikt: there is extra space, may be you could remove it also for the sake of consistency ?

src/Common/ValuesInterface.php
```diff
@@ -24,10 +24,11 @@
      *
      * @param string $col The column name.
      *
-     * @return $this
+     * @param array  $value
```
harikt: Here also.

src/Common/Update.php
pavarnos: i looked at the code and some `@params` had a blank line between, some did not. What is your preference? I'll fix the others too...

harikt: I think one space is enough for now. I was also looking at the old code.

pavarnos: still not quite understanding: should `@param` have one blank line between each declaration, or should they be on consecutive lines with no blank line between them

pavarnos: eg see https://github.com/auraphp/Aura.SqlQuery/blob/2.x/src/Common/Update.php#L83 vs https://github.com/auraphp/Aura.SqlQuery/blob/2.x/src/Common/Update.php#L153

harikt: @pavarnos I was talking about space `@param array  $value` vs `@param array $value` . I missed to understand you were talking about extra line in b/w @param. You can have an extra line in b/w @param or we can leave that for now :) .

pavarnos: ah. silly me. fixed now (3 places)

src/Common/Update.php
```diff
@@ -119,12 +119,13 @@ public function orWhere($cond)
...
-     * @return $this
+     * @param array $value
      *
+     * @return $this
      */
-    public function col($col)
+    public function col($col, ...$value)
     {
-        return call_user_func_array(array($this, 'addCol'), func_get_args());
+        return empty($value) ? $this->addCol($col) : $this->addCol($col, $value[0]);
```
harikt: I like this may be more performance ?

pavarnos: don't know. i wanted to make the function signature more explicit so it was easier for IDEs to auto-fill / auto-complete. There is another trickier one in Update.php https://github.com/auraphp/Aura.SqlQuery/blob/2.x/src/Common/Update.php#L89 which i will do if this PR gets merged. 
Got a few other PRs planned eg a fix for #100 but will start getting merge conflicts if i make too many new PRs before the current ones get merged or rejected...

src/Common/Insert.php
```diff
@@ -160,14 +160,14 @@ public function getLastInsertIdName($col)
...
+     * @param array $value Optional: a value to bind to the placeholder.
      *
      * @return $this
      *
      */
-    public function col($col)
+    public function col($col, ...$value)
     {
-        return call_user_func_array(array($this, 'addCol'), func_get_args());
+        return empty($value) ? $this->addCol($col) : $this->addCol($col, $value[0]);
```
harikt: Actually can't we just use `$this->addCol($col, ...$params);` here? Tried https://3v4l.org/sJ6S0 .

src/Common/Update.php
harikt: Actually you can just use `$this->addCol($col, ...$params);` . I have tried https://3v4l.org/sJ6S0 . I can merge some of the PR's . The interface changes probably need more review from @pmjones .

pavarnos: that is much cleaner! thanks!

harikt:
Thank you.

PR #112, opened 2016-12-30
add missing public methods to SelectInterface and related.

pavarnos:
make reset() public and add isDistinct() to round out the interface.

### DIFF
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -199,6 +199,18 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
 
     /**
      *
+     * Is the select DISTINCT?
+     *
+     * @return bool
+     *
+     */
+    public function isDistinct()
+    {
+        return $this->hasFlag('DISTINCT');
+    }
+
+    /**
+     *
      * Adds columns to the query.
      *
      * Multiple calls to cols() will append to the list of columns, not
@@ -747,7 +759,7 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
      * @return null
      *
      */
-    protected function reset()
+    public function reset()
     {
         $this->resetFlags();
         $this->resetCols();
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -307,6 +307,20 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
 
     /**
      *
+     * Has the column or alias been added to the query?
+     *
+     * @param string $alias The column or alias to look for
+     *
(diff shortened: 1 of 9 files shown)


src/Common/SelectInterface.php
```diff
@@ -79,6 +88,35 @@ public function cols(array $cols);
...
+    public function removeCol($alias);
+
+    /**
+     *
+     * Does the query have any columns in it?
+     *
+     * @return bool
+     *
+     */
+    public function hasCols();
```
harikt, line 118: I have a problem with this `hasCols` , it sounds like to check if there is a column value exists. But this looks like a check if the columns is empty or not.

harikt:
I will leave this to @pmjones for further review.

src/Common/SelectInterface.php
pavarnos: yes. i can see why it was added (so you can check if that part of the query is valid). What about adding `hasCol($name_or_alias)`?

pavarnos:
should we do UpdateInterface and InsertInterface here? The changes are mostly pretty simple (copying the declaration from the real class to the interface declaration). So they should be easy to review as one PR. What do you think?

harikt:
I feel separate may help. The benefit is if something is likely not going to merge can be rejected other PR can get merged.

src/Common/SelectInterface.php
harikt: I am ok with that :) .

src/Common/Select.php
```diff
@@ -307,6 +307,20 @@ public function removeCol($alias)
...
+     * Has the column or alias been added to the query?
+     *
+     * @param string $alias The column or alias to look for
+     *
+     * @return bool
+     *
+     */
+    public function hasCol($alias)
+    {
+        return isset($this->cols[$alias]) || array_search($alias, $this->cols) !== false;
```
harikt, line 319: Why do we need to do a search? Doesn't the isset itself help us ?

pavarnos: see https://github.com/auraphp/Aura.SqlQuery/pull/112/files/9fb886d8955663988cdfb5c371416ccdc7cdbe27#diff-1e1715bbfd6b71bb485e70bb72002475R99. cols() can add a column or an alias. This finds both, similar to how removeCol() works.

pavarnos:
this is (hopefully) the last of the changes needed to fill out the *Interface classes. 
I'd originally planned to do this as 2 separate PRs, but found that the changes for the second PR were so small it wasn't worth it So they are all together here.

PR #114, opened 2017-01-01
make $bind parameters explicit in WhereInterface and in having() clauses

pavarnos:
replaces func_get_args() with variable argument lists

### DIFF
--- a/src/Common/Select.php
+++ b/src/Common/Select.php
@@ -612,12 +612,14 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
      *
      * @param string $cond The HAVING condition.
      *
+     * @param array ...$bind arguments to bind to placeholders
+     *
      * @return $this
      *
      */
-    public function having($cond)
+    public function having($cond, ...$bind)
     {
-        $this->addClauseCondWithBind('having', 'AND', func_get_args());
+        $this->addClauseCondWithBind('having', 'AND', $cond, $bind);
         return $this;
     }
 
@@ -629,14 +631,16 @@ class Select extends AbstractQuery implements SelectInterface, SubselectInterfac
      *
      * @param string $cond The HAVING condition.
      *
+     * @param array ...$bind arguments to bind to placeholders
+     *
      * @return $this
      *
      * @see having()
      *
      */
-    public function orHaving($cond)
+    public function orHaving($cond, ...$bind)
     {
-        $this->addClauseCondWithBind('having', 'OR', func_get_args());
+        $this->addClauseCondWithBind('having', 'OR', $cond, $bind);
         return $this;
     }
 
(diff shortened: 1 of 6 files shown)


harikt:
Thank you.

src/Common/Select.php
```diff
@@ -612,12 +612,14 @@ public function groupBy(array $spec)
      *
      * @param string $cond The HAVING condition.
      *
+     * @param array ...$bind arguments to bind to placeholders
```
harikt, line 615: I have a problem with the doc type. We may want to refer how the PSR-5 is defining..

```php
function query1($hello, ...$args)
{
    var_dump($args);
}

query1('Hello', 'World', 'And');
query1('Hello', ['World', 'And']);
```

For the `$args` we are mentioning it is an array. ie correct, but will it confuse people and send an array as in the second example?

See https://3v4l.org/g8Wgm

I am not seeing anything quick on my search : https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

pavarnos: ... me either. might be a useful issue to raise with them? i agree it is confusing. Is a better alternative `mixed`? Or maybe its better to clarify the documentation a  little.

pavarnos: another thought: the code for WhereInterface is duplicated across 3 classes. Would it be better to extract 'where()` and `orWhere()` into a WhereTrait to remove the duplication and simplify maintenance?

harikt: @pavarnos I don't know how you are planning to do the same for the trait. If I can see some code it may help. Yes we could probably ask someone on twitter ;) .

harikt: @pavarnos according to my understanding in this case this should be string than array. Tested with https://3v4l.org/WvpMR . Also from twitter https://twitter.com/localheinz/status/815468361463332864 I was looking at http://stackoverflow.com/questions/13787426/how-to-declare-unlimited-parameters-in-docblock/24337184#24337184 .

pavarnos: OK. That makes sense. Declaring `@param string...$bind` is no problem (phpStorm complains, but it seems to be lagging behind the standard so I'm not too worried).
If we declare `function addCol($col, string ...$value)` then the minimum php version is 7.0. Is that OK?
Moving to 7.0 would let us do some other good stuff like `declare(strict_types=1)`

pavarnos: it can't be `string` because it may not always be a string. https://github.com/auraphp/Aura.SqlQuery/blob/3.x/src/AbstractQuery.php#L403. Is `mixed` better than a union type eg `string|SubselectInterface`?

harikt: I was not suggesting to use php 7 string. It was just an example. But I don't have any trouble if we move to 7.0 as 5.6 has EOL . May be another issue :) .

pavarnos: OK. so they should all be `mixed`? (which allows string and SubselectInterface)

harikt: Interesting question. Mixed doesn't actually make sense for it can also be array and any type. May be `string|SubselectInterface` ? We can create an issue for discussion. So others can join ?

pavarnos: sounds good. http://docs.php.net/manual/en/language.pseudo-types.php

> mixed indicates that a parameter may accept multiple (but not necessarily all) types

PR #115, opened 2017-01-01
add Insert::orReplace

pavarnos:
originally added a useReplace() as suggested by #100, but then found sqlite has an orReplace() function and it will be very handy to have the same signature for both dbms when writing unit tests: can switch out mysql for sqlite and not have to add code to check which you are using.

### DIFF
--- a/tests/Mysql/InsertTest.php
+++ b/tests/Mysql/InsertTest.php
@@ -58,6 +58,21 @@ class InsertTest extends Common\InsertTest
         $this->assertSameSql($expect, $actual);
     }
 
+    public function testOrReplace()
+    {
+        $this->query->orReplace()
+                    ->into('t1')
+                    ->cols(array('c1', 'c2', 'c3'))
+                    ->set('c4', 'NOW()')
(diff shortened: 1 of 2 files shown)


tests/Mysql/InsertTest.php
```diff
@@ -58,6 +58,21 @@ public function testHighPriority()
...
+        $this->query->orReplace()
+                    ->into('t1')
+                    ->cols(array('c1', 'c2', 'c3'))
+                    ->set('c4', 'NOW()')
+                    ->set('c5', null);
+
+        $actual = $this->query->__toString();
+        $expect = sprintf($this->expected_sql_with_flag, '');
+        $expect = preg_replace('|INSERT |ms', 'REPLACE', $expect);
+
```
harikt: It would have been good if you could write the raw query itself than using a preg_replace. What I mean is something as https://github.com/auraphp/Aura.SqlQuery/pull/115/files#diff-f685026c17cbf1227a63eb3b344b96a4R27 .

PR #116, opened 2017-01-01
move copy/pasted where() / orWhere() to WhereTrait

pavarnos:
could do similar for LimitInterface, LimitOffsetInterface and ValuesInterface.

### DIFF
--- a/src/Common/Delete.php
+++ b/src/Common/Delete.php
@@ -19,6 +19,8 @@ use Aura\SqlQuery\AbstractDmlQuery;
  */
 class Delete extends AbstractDmlQuery implements DeleteInterface
 {
+    use WhereTrait;
+
     /**
      *
      * The table to delete from.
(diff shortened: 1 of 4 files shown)


harikt:
@pavarnos I have a question here. 

Do we really need trait or can we just add these methods to [AbstractQuery](https://github.com/auraphp/Aura.SqlQuery/blob/2.x/src/AbstractQuery.php) class ?

pavarnos:
Could do. But where() / orWhere() is not used by Insert so the methods do not make any sense on that class. So then we'd need to have some way to hide them from the  public interface of the Insert class?
Similarly with limit/offset. 
ValuesInterface is used by Insert and Update (and potentially by Select) but not by Delete. So there is no obvious hierarchy.

Its just an idea. Delete the PR if you think it is going in the wrong direction.

harikt:
Sorry I didn't thought more about that :) .

pavarnos:
could also do similar for ReturningInterface, which is only used by pgsql... but yeah, how far should we go?

PR #117, opened 2017-01-01
fix scrutinizer messages

pavarnos:


harikt:
scrutinizer says coverage reduced. May be you could look into it / add some more test to increase the coverage?

PR #120, opened 2017-01-02
move copy/pasted limit()/offset() code to traits

pavarnos:
similar to #116: the code for limit() and offset() was duplicated in several classes. Extracted the common code to traits, added getLimit() and getOffset() to the relevant interfaces.

### DIFF
--- a/src/AbstractQuery.php
+++ b/src/AbstractQuery.php
@@ -459,20 +459,4 @@ abstract class AbstractQuery
 
         return PHP_EOL . 'ORDER BY' . $this->indentCsv($this->order_by);
     }
-
-    /**
-     *
-     * Template method overridden for queries that allow LIMIT and OFFSET.
-     *
-     * Builds the `LIMIT ... OFFSET` clause of the statement.
-     *
-     * Note that this will allow OFFSET values with a LIMIT.
-     *
-     * @return string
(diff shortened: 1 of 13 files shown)


pavarnos:
This PR is not as smooth and simple as #116 because Common\Select::page/paging has complex interactions with limit/offset. 
But: note the improvement in code quality: Scrutinizer reports 9.34 (up from 8.8?), mostly influenced by Common\Select

src/AbstractQuery.php
```diff
@@ -491,22 +473,6 @@ protected function buildOrderBy()
      */
     protected function buildLimit()
     {
-        $clause = '';
```
harikt, line 494: if this method just returns an empty string what is the value for it ?

pavarnos: as it says in the comment: it is a template method https://en.wikipedia.org/wiki/Template_method_pattern "one or more algorithm steps can be overridden by subclasses to allow differing behaviors while ensuring that the overarching algorithm is still followed"
The buildLimit() is here to hold its place in the build() algorithm for a derived method to do something useful with.

pavarnos: It is needed by Delete and Update

PR #123, opened 2017-01-17
add “composer test” command to composer.json

pavarnos:
The phpunit test setup in https://github.com/pmjones/psr7immutability is very slick. Added it here.


pmjones:
Accepting this provisionally, and reserve the right to change back to `./vendor/bin/phpunit`.

PR #127, opened 2017-02-18
Splitting array parameters into multiple parametres for fix PDO excecution error "Array to string conversion"

PHPCraftdream:


harikt:
You probably need Aura.Sql here to execute the `IN` statements.

pmjones:
@harikt Believe it or not, this is something I've been contemplating, at least for the 3.x version. Maybe it should be in the 2.x version as well.

PHPCraftdream:
@harikt I used Aura\Sql\ExtendedPdo object to execute sql query with array when I have recieved error "Array to string conversion". It seem there is no way to make it without these fixes.

pmjones:
@PHPCraftdream Can you do me a favor and post the code that generated the error? I'd like to test it myself and see what's going on.

PHPCraftdream:
@pmjones 
```PHP
	error_reporting(E_ALL);

	$pdo = new \Aura\Sql\ExtendedPdo(
		'mysql:host=localhost;dbname=test',
		'test',
		'test'
	);

	$queryFactory = new \Aura\SqlQuery\QueryFactory('mysql');

	$selectQuery = $queryFactory->newSelect();
	$selectQuery->cols(['*'])->from('session')->where('`id` in (?)', [1, 3]);	

	$statement = $pdo->prepare($selectQuery->getStatement());
	$res = $statement->execute($selectQuery->getBindValues());
````

**Comment by pmjones** [comment]
The question now is, does this represent a BC break? I can imagine people writing tests that examine the statement, expecting a single placeholder instead of multiples.

Would it be better to put this on the 3.x branch instead?

**Comment by harikt** [comment]
@pmjones this is a bug for it was not working. So this is never a BC break. If this bug persists probably no one would have been using the feature or never complained about the same.

**Comment by pmjones** [comment]
@PHPCraftdream or @harikt can one of you make a similar PR against the 3.x branch?

**Comment by harikt** [comment]
@pmjones @PHPCraftdream I have been trying to send a PR to 3.x .

In the mean time I was testing out things and noticed there is no problem with Aura.Sql or Aura.Sqlquery part . Wonder whether this fix is really needed also. 

Reason : In this case we are forced to write each placeholders for the ` IN (?) ` .

This is why it went wrong as reported, you were using normal pdo statements. But instead you should have used perform method.

@pmjones we probably need to revert back and bring something different in 3.x .

```php
<?php
require __DIR__ . '/vendor/autoload.php';

$query_factory = new Aura\SqlQuery\QueryFactory('mysql');
$select1 = $query_factory->newSelect();
$select1->from('articles')
    ->cols(['*'])
    ->where('id IN (:id)')
    ->bindValue('id', [1, 2, 3, 4]);
echo $select1->__toString();
var_dump($select1->getBindValues());

$pdo = new Aura\Sql\ExtendedPdo(
    'mysql:host=localhost;dbname=db',
    'root',
    'root'
);
$sth = $pdo->perform($select1->__toString(), $select1->getBindValues());
echo $sth->queryString;

$result = $pdo->fetchAll($select1->__toString(), $select1->getBindValues());

$select2 = $query_factory->newSelect();
$select2->from('articles')
    ->cols(['*'])
    ->where('id IN (?)', [1, 2, 3, 4]);
echo $select2->__toString();
var_dump($select2->getBindValues());

$sth = $pdo->perform($select2->__toString(), $select2->getBindValues());
echo $sth->queryString;

$result = $pdo->fetchAll($select2->__toString(), $select2->getBindValues());
```

Sorry for not looking more carefully into this.

PHPCraftdream:
May be it should be fixed with interfaces like this:

```PHP
     if ($bind_value instanceof MultipleParametresInterface) {
          $bind_value_arr = $bind_value->getArray();
          //.........
          foreach ($bind_value_arr as $subValue) {
               //.........
          }
     }

```
Instead:
```PHP
     if (is_array($bind_value) && !empty($bind_value)) {
          //.........
          foreach ($bind_value as $subValue) {
               //.........
          }
     }
```
I can remake fixes with interfaces, but right now it is already pretty useful

PR #131, opened 2017-03-18
Extract builder logic to separate classes

pmjones:
To separate the concerns of "collecting" query parts, and "building" them into a statement, this PR moves the `build*()` methods to their own classes.  This entails creating a builder for each kind of statement (select/insert/update/delete) and in some cases a driver-specific builder for the statement (mysql insert and pgsql everything).

@pavarnos @harikt etc: any comments or concerns?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ php:
   - hhvm
   - 7
 script:
-  - phpunit --coverage-clover=coverage.clover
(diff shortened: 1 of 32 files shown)


harikt:
Quick look, I didn't find any problem. It looks all api and tests are working without changes. So I am :+1:  .

pavarnos:
Nice

Issue #108, opened 2016-09-20
Version 3: Update interfaces and nothing else


pmjones:
There have been several questions lately regarding the fact that the interfaces do not match the classes. The disconnect is because the classes have had features added, but adding the relevant methods to the interface would be a BC break.

Perhaps the time has come to update the interfaces, and in so doing move to version 3, with no other functionality changes.

Thoughts?

cdekok:
If you plan to change it you might want to consider changing the split with SubSelectInterface and SelectInterface it's a bit annoying you cannot get any typehint as you can only typehint on one of interfaces please consider extending the interface for ease of use with an IDE.

pavarnos:
Yes please!
Also fix col() which is yellow in phpStorm with 2 arguments.

I have ended up creating a small wrapper class that fixes some annoyances and adds extra utility functions. For example, adding the ability to include columns from the join table that get automatically prefixed by the table name, a joinLeft() / joinInner(), and extra methods to ease working with subselects (not shown below). 

```php
class WrappedSelect {
   /** @var Aura\SqlQuery\Common\SelectInterface */
   private $select;

    /**
     * Adds a JOIN table and columns to the query.
     * @param string       $joinType  The join type: inner, left, natural, etc.
     * @param string       $spec      The table specification; "real_table" or "real_table AS alias" 
     * @param string       $condition Join on this condition.
     * @param array        $columns   columns as passed to cols()
     * @return $this
     */
    public function join($joinType, $spec, $condition = null, $columns = null)
    {
        $this->select->join($joinType, $spec, $condition);
        $this->addColumns($columns, $spec);
        return $this;
    }

    /**
     * @param string $spec      The table specification; "foo" or "foo AS bar".
     * @param string $condition Join on this condition.
     * @param array  $columns   columns as passed to cols()
     * @return $this
     */
    public function joinLeft($spec, $condition = null, $columns = [])
    {
        return $this->join('left', $spec, $condition, $columns);
    }

    /**
     * Adds a FROM element to the query; quotes the table name automatically.
     * @param string $table The table specification; "foo" or "foo AS bar".
     * @param array  $columns
     * @return $this
     */
    public function from($table, $columns = [])
    {
        $this->select->from($table);
        $this->addColumns($columns, $table);
        return $this;
    }

    /**
     * @param array  $columns
     * @param string $tableName table name
     */
    private function addColumns($columns, $tableName)
    {
        if (empty($columns)) {
            return;
        }
[...]

PR #132, opened 2017-03-19
3.x quoter

pmjones:
This PR creates separate quoter classes for each driver (notably mysql and sqlsrv).

@harikt @pavarnos Thoughts?


harikt:
@pmjones 

Thank you for pinging.

I am not sure what advantage do you see in moving to a separate class than passing the string?

pavarnos:
@harikt it removes db specific config from the generic QueryFactory class https://github.com/auraphp/Aura.SqlQuery/pull/132/files#diff-f4d2c86a7d03f40d9533ddaca828a28fL47 into the db specific driver directory, which is an improvement in clarity. It makes the quoter re-usable too (eg in Aura.Sql?).

@pmjones how is the test coverage? I can't seem to find it...

pmjones:
@harikt As @pavarnos notes, it's "better form" than before, but not in any way functionally critical.

@pavarnos Test is at https://github.com/auraphp/Aura.SqlQuery/pull/132/files#diff-c13b8f7f14b2d2fbc0552d98128d064f and coverage should be the same as before (i.e. 100%); the db-specific quoters only modify properties, not methods. Cf. https://scrutinizer-ci.com/g/auraphp/Aura.SqlQuery/?branch=3.x

Other thoughts?

harikt:
@pmjones @pavarnos thank you.

The quoter is actually introduced in https://github.com/auraphp/Aura.Sql/pull/153 .

So I doubt about it.

pavarnos:
Thanks. Temporary blindness: I cannot find the coverage report / percentage?

pmjones:
@harikt So is that a "yes" or a "no" on this PR?

@pavarnos At the Scrutinizer link, on the right, under "Badges", it shows 100%.

harikt:
@pmjones I was answering to @pavarnos 

> It makes the quoter re-usable too (eg in Aura.Sql?).

The quoter is actually introduced in auraphp/Aura.Sql#153 .

So I doubt about it.

Probably the Quoter in that case can be moved out. But that may break what we advertise, but less things being duplicated.

pmjones:
@harikt Ah right. Yes, the quoter in Sql is much more limited.

@harikt @pavarnos Just so that we can either merge or close this PR, and I back to the "actual plan" laid out in #108 , give me "go" or "no-go" and I'll work from there.

harikt:
@pmjones I am :+1:  . So "go" .

pavarnos:
good with me

pmjones:
Thanks folks. Back to #108 !

PR #133, opened 2017-03-20
Remove SubSelect interface, typehint on SelectInterface instead

pmjones:
Per my comment on #108, this PR typehints binding of sub-select statements on SelectInterface, not SubselectInterface (which was originally introduced as a BC-mitigator).


pmjones:
@harikt @pavarnos @auroraeosrose Please comment as convenient!

harikt:
Probably when breaking changes are happening, it may be good to add in changelog also. So we don't forget or want to do look for the changes.

Issue #129, opened 2017-03-03
Postgre jsonb operators and placeholders


taxp:
Hello!

I have a problem with making select query with operator ?| that is used in postgre to check if two jsonb items have common keys. 
My query, written by hand and simplified for the example, is:

```sql
select * 
from table_name
where phones::jsonb ?| ARRAY['71111111111', '71234567890']
```

I used query builder like this:
```php
$queryFactory = new QueryFactory('pgsql');
$select = $queryFactory->newSelect();

$select->cols(['*'])
->from('table_name')
->where("phones::jsonb ?| ARRAY['71111111111', '71234567890']");

$statement = $select->getStatement();
```

but I got that (notice how "?" has been replaced. Builder supposes that it is placeholder for parameter):
```sql
select * 
from table_name
where phones::jsonb :_1_| ARRAY['71111111111', '71234567890']
```

So, what can I do to get rid off of this?

pavarnos:
should this be a bug report on https://github.com/auraphp/Aura.Sql ?

taxp:
@pavarnos , no, it is about the query builder. Look, I found that every "where" condition leads to AbstractQuery's addWhere method and, in the end, to rebuildCondAndBindValues. So there is used regexp to split query by question sign and that causes the noted problem.

pmjones:
@taxp Would it be sufficient to add a "rawWhere()" method that take the WHERE clause as-is and does not attempt to bind data data to it at all?

taxp:
@pmjones, yes, I guess that can solve the issue. Also, it is need to notice that "raw where" methods should implement both andRawWhere and orRawWhere options.

pavarnos:
since we are allowed BC breaks, I wonder if a cleaner solution might be to split off the parameter substitution...
eg remove the ...$bind (for ? parameter substitution) from `where()` to simplify its interface and move that functionality to `whereWithBoundParameters()` (or similar).

pmjones:
@pavarnos -- Not bad! If we're doing BC breaks of that kind, though, we're off plan #108 (which is fine with me at this point). Further, maybe the thing to do is eliminate support for ?-mark placeholders entirely, which might *really* simplify things. The 2nd arg on where() et al. can become a key-value array that just calls bindValue() right then. E.g.:

```
$select->where('foo = :foo', ['foo' => 'foo_value']);
// behind the scenes, it calls $select->bindValues(['foo' => 'foo_value'])
```

[...]

PR #134, opened 2017-03-21
3.x: Named placeholders only for where() et al. conditions

pmjones:
In reference to #129 , this PR removes support for ?-placeholders in where(), having(), and join*() methods. You can still bind at the time of the method call, but you must specify the named placeholder to bind to. E.g., instead of this ...

```
$select->where('id = ?', $id_value);
```

... you must do this:

```
$select->where('id = :id', ['id' => $id_value]);
```

Support for sub-selects as bound values is retained.

This is obviously more typing, but is more explicit, and requires less internal calculating of placeholder names.

@harikt @taxp @pavarnos et al: let me know what you think.


pavarnos:
Looking good: a lot of complicated code has been deleted, which is always a reason to celebrate (can i buy you a beer?)

Is it worthwhile / useful to this to its extreme to further simplify the interface and suggest that the only place you can bind values is in bindValue() / bindValues()? It is kinda convenient being able to where() and bind all in the same function call.

https://github.com/auraphp/Aura.SqlQuery/blob/3.x-named-placeholders-only/src/AbstractQuery.php#L278 is now incorrect? Maybe grep for 'placeholder' to see if there are others?

pmjones:
> Is it worthwhile / useful to this to its extreme to further simplify the interface and suggest that the only place you can bind values is in bindValue() / bindValues()?

That would make it more difficult to replace :placeholder with a SELECT object in the condition. However, there might be a way to do so at build() time, rather than where() et al. time. For now, I'm satisfied to leave it in place.

> Maybe grep for 'placeholder' to see if there are others?

Good catch. I need to walk through `docs/` again anyway, and can attend to it then.

harikt:
I am good with this.

One thing probably I missed to see is a test where the same name place holder can be replaced at multiple places.

```
SELECT * FROM table WHERE id = :foo AND foo = :foo 
```

Do we have a test that cover this already ? May be quick look in diff I probably missed.

harikt:
@pmjones can we delete the branch also ?

pmjones:
> [Can] the same name place holder can be replaced at multiple places [?]

Not in Aura.SqlQuery logic itself, no -- that has to happen at the Aura.Sql level for now. I anticipate a future change where some of the Aura.Sql "rebuilder" logic gets incorporated into Aura.SqlQuery.

pmjones:
> can we delete the branch also ?

It's been, like, 10 minutes -- give a brother a some breathing room. ;-)

Issue #121, opened 2017-01-12
Group by [Feature]


harikt:
Hi, 

One of the missing feature I think about the library is creating grouping statements. I have been looking at https://github.com/paragonie/easydb#grouping-of-conditions .

An example taken : 

```php
$statement = EasyStatement::open()
    ->group()
        ->with('subtotal > ?')
        ->andWith('taxes > ?')
    ->end()
    ->orGroup()
        ->with('cost > ?')
        ->andWith('cancelled = 1')
    ->end();

echo $statement; /* (subtotal > ? AND taxes > ?) OR (cost > ? AND cancelled = 1) */
```

Does anyone like this feature ?

pavarnos:
would be nice to do a similar syntax with where()

pmjones:
I am not enthusiastic about this. It can be accomplished just as easily with 2.x or 3.x, either like so ...

```php
$select
    ->where(' ( subtotal > :subtotal')
    ->where('taxes > :taxes ) ')
    ->orWhere(' ( cost > :cost')
    ->where('cancelled = 1 ) ');
```

... or like so:

```php
$select
    ->where(' ( subtotal > :subtotal AND taxes > :taxes ) ')
    ->orWhere(' ( cost > :cost AND cancelled = 1 ) ');
```

Do you see what I mean?

harikt:
I see what you mean, but considering the use of query builder 

Example

```sql
$query->group()
if ($this->subtotal) {
    $query->where('subtotal = :subtotal', $this->subtotal)
}
if ($this->taxes) {
    $query->where('taxes = :taxes', $this->taxes)
}
$query->end()
```

UPDATE : This query probably will be wrong when the subtotal and taxes are null .

pmjones:
/me tilts head

OK, maybe.

At the very least, `group()` is too easily confused with `groupBy()` so you'd need a different name. Maybe `begin()` or `open()` or `start()` or something like that.

Further, I imagine that this is a relatively complex thing to implement. It will need ...

- At least 4 new methods (`begin()` and `end()` methods for both `where()` and `having()`)
- Internal tracking to see if the where/having stack is "in a group" or not
- Automatic `end()`-ing at `build()` time to close groups
- A way to check if the "group" is empty, so that the builder doesn't place `( )` (empty parentheses) in the statement if there are no group elements

I'm sure there is even more to it than that.

harikt:
@pmjones I agree with you, it is a bit complex.

[...]

PR #136, opened 2017-03-22
Support 'grouping' of where() and having() conditions

pmjones:
This allows for `where(function ($select) {...} )`, and put parens around the resulting conditions.

References #121 .


harikt:
Brilliantly done. One two emojis missing which we don't need for this.

Great work!. 

Thank you for your efforts.

harikt:
Oh may be we need a test for having also ?

pmjones:
> Brilliantly done.

You're kind. :-)

> Oh may be we need a test for having also ?

I was going to say "both where() and having() lead to the same methods" -- but then, it can't hurt, so I'll add one.

PR #141, opened 2017-04-18
Only fall back to common quoter when DB-specific quoter doesn't exist

### LINKED ISSUE #138: Common quoter is always used when using common query factory
See here: https://github.com/auraphp/Aura.SqlQuery/blob/3.x/src/QueryFactory.php#L209

This results in the common quoter being used when generating MySQL queries, which results in the incorrect quote type being used. Removing the `$this->common` check resolves the issue.

I believe this came about as part of the refactoring for https://github.com/auraphp/Aura.SqlQuery/pull/132.

djmattyg007:
Fixes #138


pmjones:
Thanks for this!

PR #147, opened 2017-07-03
Drop HHVM support

odan:
> HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.
https://travis-ci.org/auraphp/Aura.SqlQuery/jobs/249719786

PR #148, opened 2017-07-21
Make travis use trusty

jakejohns:
#147 suggests dropping hhvm support. I'm not sure why we'd do that, 
although I see [that was done in Aura/Auth](https://github.com/auraphp/Aura.Auth/commit/811f799aca9df07bf73c1c3a21221dfc4ff4b2c9) seemingly without explanation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: php
+dist: trusty
 php:
   - 5.6
   - hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.6
   - hhvm
   - 7
+  - 7.1
 script:
   - composer install
   - ./vendor/bin/phpunit --coverage-clover=coverage.clover


jakejohns:
although I guess [this won't be required after Friday, September 1 2017 ](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming)

jakejohns:
*bump* @pmjones , @harikt any thought on this or #147

.travis.yml
```diff
@@ -1,5 +1,6 @@
 sudo: false
 language: php
+dist: trusty
```
harikt, line 3: can you add php 7.1 also ?

harikt:
Happy to merge. And I didn't noticed the PR waiting for so long. Sorry about that.

Please add php 7.1 also to travis. I will merge once done. Ping me if I didn't noticed.

harikt:
I think the hhvm is not going to work, may be we should consider #147 itself :) .

PR #154, opened 2017-11-29
Drop HHVM support v2.x

odan:


odan:
The build for PHP 5.3 failed on Travis because of this error message:

> PHP 5.3 is supported only on Precise.
See https://docs.travis-ci.com/user/reference/trusty#PHP-images on how to test PHP 5.3 on Precise.
Terminating.

Should we drop the support for PHP 5.3 in the v2.x branch?

harikt:
We could use something like this : https://github.com/auraphp/Aura.Auth/blob/99c16a19f537dfb9fc33e5c28dd61681a9a2fa3d/.travis.yml#L5 , so 5.3 will run I believe.

PR #156, opened 2018-01-22
Db platform constants

simPod:
I'd like to reference those constants in my project, it's more safe as constants than raw strings

Also, removed unnecessary complexity from QueryFactory constructor

### DIFF
--- a/src/QueryFactory.php
+++ b/src/QueryFactory.php
@@ -17,10 +17,11 @@ namespace Aura\SqlQuery;
  */
 class QueryFactory
 {
-    /**
-     * Use the 'common' driver instead of a database-specific one.
-     */
     const COMMON = 'common';
+    const MYSQL = 'Mysql';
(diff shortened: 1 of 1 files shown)


harikt:
Why not extend the `QueryFactory` ? Is there any problem with it ? As these constants are not used by Aura.SqlQuery itself I wonder whether it is really needed here.

harikt:
I also noticed you send this https://github.com/doctrine/dbal/pull/2986 to doctrine. Is there any specific thing you are looking from these PR's ?

simPod:
Well, I think there would be no point in extending. I could specify those constants somewhere aside but I was thinking it might be useful to add them to these two packages. Especially here, I think it's better to pass constant with platform name rather than simple string.

I'm mapping this package to DBAL so I need to create a map using those constants, not simple strings if possible. However, my current solution uses strings only and waits if these will be merged or not :)

PR #162, opened 2018-04-03
Support inline array for condition

syrm:
With this PR you can inline array :
`->where('user_id IN (?)', [[1, 2, 3]])`
Or like this : 
`->where('user_id IN (:ids)', ['ids' => [1, 2, 3]])`

### DIFF
--- a/tests/Common/SelectTest.php
+++ b/tests/Common/SelectTest.php
@@ -509,6 +509,39 @@ class SelectTest extends AbstractQueryTest
         $this->assertSame($expect, $actual);
     }
 
+    public function testWhereWithInlineArray()
+    {
+        $this->query->cols(array('*'));
+        $this->query->where('c1 = c2')
+            ->where('c3 = :c3 AND c4 IN (:c4) AND c5 = :c5', ['c3' => 'foo', 'c4' => [3, 2, 1], 'c5' => 'bar'])
+            ->where('c6 = ? AND c7 IN (?)', ['foo1', [6, 5, 4]]);
+        $expect = '
+            SELECT
+                *
+            WHERE
+                c1 = c2
+                AND c3 = :c3 AND c4 IN (:__1__, :__2__, :__3__) AND c5 = :c5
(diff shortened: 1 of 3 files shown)


tests/Common/SelectTest.php
```diff
@@ -712,7 +745,7 @@ public function testUnionAll()
     public function testAutobind()
     {
         // do these out of order
-        $this->query->having('baz IN (:baz)', ['baz' => ['dib', 'zim', 'gir']]);
+        $this->query->having('baz IN (?, ?, ?)', ['dib', 'zim', 'gir']);
```
harikt, line 748: Why did you changed this test ? Can you add it as a separate test ? Same for below. So we can make sure nothing is breaking. So far I like the PR. Nice.

syrm: Because this test was wrong and now it don't pass with my modification.
And there is a new test for my code : testWhereWithInlineArray

jimbo8098:
Would be great to see this added in, looks good to me!

tests/Common/SelectTest.php
koriym: @harikt I think @syrm is right. The test is probably wrong. Take a look here.

https://github.com/auraphp/Aura.SqlQuery/blob/091f470e6d8ebbaafd978a7984b231a4f8279ecc/tests/Common/SelectTest.php#L731

It is not possible to bind such an array.

koriym (commented):
Looks good to me. (Thanks to @NaokiTsuchiya for joining the review.)

tests/Common/SelectTest.php
koriym: Awesome PR thanks @syrm

koriym:
I've prepared a refactored PR for this in https://github.com/auraphp/Aura.SqlQuery/pull/198.

harikt:
Hey @koriym , 

Can we merge this PR and make the necessary changes on the other?

The other one looks like the commit is not getting the correct user.

harikt:
Thank you @syrm . I am sorry that this PR waited so long.

Issue #172, opened 2020-01-13
To add IGNORE statement to INSERT query library has different interfaces


maximTarleckiy:
To add IGNORE statement to INSERT query library has different driver implementations:
\Aura\SqlQuery\Pgsql\Insert:ignore()
\Aura\SqlQuery\Sqlite\Insert::orIgnore()
Can we use common interface for this?
Like: \Aura\SqlQuery\Common\Insert::ignore()

PR #173, opened 2020-01-16
Add ignore() method to common Insert class

maximTarleckiy:
https://github.com/auraphp/Aura.SqlQuery/issues/172

### DIFF
--- a/src/Common/Insert.php
+++ b/src/Common/Insert.php
@@ -273,6 +273,21 @@ class Insert extends AbstractDmlQuery implements InsertInterface
 
     /**
      *
+     * Adds IGNORE flag depending on DB syntax.
+     *
+     * @param bool $enable Set or unset flag (default true).
+     * @throws Exception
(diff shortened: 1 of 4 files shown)


harikt:
I agree with you. It looks `orIgnore` seems to be used inside update also.

Only question here is do we want to use `$inser->ignore()` or `$inser->orIgnore()`.

koriym (approved):

NaokiTsuchiya:
LGTM

harikt:
@koriym , @NaokiTsuchiya  please do make sure other methods are also changed.

koriym:
@harikt Sorry we missed it.

harikt:
As we move to 3.x BC is not an issue. But I wish we could address using ignore or ignoreOr

koriym:
@harikt `orIgnore` method marked `@deprecated`?

harikt:
That is the change in this PR. But I doubt about it.

PR #174, opened 2020-01-16
Add to composer ext-pdo_sqlite which required for tests

maximTarleckiy:
PDO SQLIte extension required for tests.


harikt:
Thank you.

PR #184, opened 2021-05-02
Add lateral joins

golgote:
Postgresql supports lateral joins but they currently cannot be used. This makes it possible.


golgote:
I don't know why build fails under Php 7.3, it doesn't look like it is related to my small addition...

golgote:
Hi, can you please merge this (and fix your travis so that php 7.3 tests are supported). Thanks

PR #186, opened 2021-06-26
Testing against PHP 8.0

AllenJB:
This PR updates the 3.x branch with testing against PHP 8.0

Due to return type compatibility issues with the setUp() method, I could find no way to maintain compatibility with PHP versions older than 7.1 without duplicating the tests.

If you want me to duplicate the tests or have any alternative ideas on how to maintain compatibility with older PHP versions, please let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,15 @@
 sudo: false
 language: php
-dist: trusty
+dist: bionic
 php:
-  - 5.6
-  - 7
   - 7.1
   - 7.2
   - 7.3
   - 7.4
+  - 8.0
 script:
   - composer install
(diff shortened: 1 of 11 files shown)


AllenJB:
Travis-CI seems to have failed to run on this PR, but I don't know why. Possibly an issue with travis-ci.org vs travis-ci.com?

I set up Travis on my fork for testing while I was working on these changes and tests passed there: https://travis-ci.com/github/AllenJB/Aura.SqlQuery/builds/230944424

harikt:
Thank you for your work. Unfortunately we cannot drop support for 5.x in 3.x. So this has been fixed via https://github.com/auraphp/Aura.SqlQuery/pull/195 . We are using yoast polyfill now.

PR #187, opened 2021-07-05
[WIP] Fix xdebug mode for Travis CI

afilina:
Error message: `Code coverage needs to be enabled in php.ini by setting 'xdebug.mode' to 'coverage'`
Solution found here: https://travis-ci.community/t/xdebug-3-is-installed-by-default-breaking-builds/10748

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ php:
   - 7.2
   - 7.3
   - 7.4
+env:
+  global:
+    - XDEBUG_MODE=coverage
 script:
   - composer install
   - ./vendor/bin/phpunit --coverage-clover=coverage.clover


afilina:
@pmjones The failure seems unrelated to my changes. I'm not sure why Scrutinizer runs _before_ the coverage is generated with PHPUnit via Travis CI.

I see very little activity in the last few years. If Scrutinizer's behavior changed over the years to the point of needing a rework of the pipeline, then perhaps we can drop it for now so that it doesn't block the Tavis CI tests.

djmattyg007:
In general I think a move away from Travis CI is sorely needed. Ideally it would be replaced by Github Actions.

afilina:
@djmattyg007 Ok, I'll give it a stab this week.

harikt:
This has been fixed via https://github.com/auraphp/Aura.SqlQuery/pull/195

PR #190, opened 2021-10-27
Correct preg_split call

srjlewis:
Since version 2.* is still the latest released version, in PHP 8.1 preg_split is creating deprections notices due to null being used.

Also in the php docs is shows the value for no limit should be -1.


harikt:
Hey @srjlewis ,

Is it possible for you to send to 3.x?

Thank you.

harikt:
Ignore my comment on 3.x . 3.x seems not using that function.

PR #197, opened 2022-02-03
Fix types in phpdoc

koriym:
The wrong type of phpdoc will cause false positive errors in static analysis such as phpstan and psalm in applications that depending this library. This PR fixes that problem.

### DIFF
--- a/src/Common/Delete.php
+++ b/src/Common/Delete.php
@@ -34,14 +34,14 @@ class Delete extends AbstractDmlQuery implements DeleteInterface
      *
      * Sets the table to delete from.
      *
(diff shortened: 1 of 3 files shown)


src/Common/Delete.php
```diff
@@ -34,14 +34,14 @@ class Delete extends AbstractDmlQuery implements DeleteInterface
      *
      * Sets the table to delete from.
      *
-     * @param string $table The table to delete from.
+     * @param string $from The table to delete from.
```
harikt, line 37: Do you want to change this variable.

koriym: It's different from the interface.

PR #198, opened 2022-02-04
Refactor #162

koriym:
The nesting of if was not easy to read, so I created a method.

See https://github.com/auraphp/Aura.SqlQuery/pull/198/commits/66333facc3ec34cb16e75676798567c9f9af48c6


harikt:
Thank you @koriym .

PR #206, opened 2023-04-20
[BUG] Fix bad use of strripos

auroraeosrose:
It is 100% valid for strripos to return 0 and still be considered a match
we want quotes in that case for the AS blah
NEVER EVER if a strripos, always explicitly check for false!!

This was breaking jsonb text returns with mixed case aliases

Select "pt"."name"->>'en_US' AS "myDescription" from mytable pt is what we want to end up with 

The replaceNamesAndAliasIn was getting ' AS myDescription' and not properly escaping it because the strripos was returning 0

### DIFF
--- a/src/Common/Quoter.php
+++ b/src/Common/Quoter.php
@@ -206,7 +206,7 @@ class Quoter implements QuoterInterface
     {
         $quoted = $this->replaceNamesIn($val);
         $pos = strripos($quoted, ' AS ');
-        if ($pos) {
+        if ($pos !== false) {
             $alias = $this->replaceName(substr($quoted, $pos + 4));
             $quoted = substr($quoted, 0, $pos) . " AS $alias";
         }


harikt:
Hey @auroraeosrose ,

Thank you for the PR.

Can I add you to one of the maintainers. So you can also merge and release soon. I didn't find this earlier or missed for some reason. So don't want PR to wait so long.

harikt:
@koriym I guess I made a mistake. We never released 4.x . So merging to 4.x was something I did wrong. I guess it should go to 3.x.

auroraeosrose:
Should probably go to both if the issue exists in both branches
I'd have to look at the code :)

harikt:
I have pushed to 3.x, 2.x . 4.x seems not released yet. 

Bug fixes are released in 2.8.1 and 3.0.0 .

Thank you.

koriym:
@harikt Yes. The 3.x test covers PHP 5.6 through PHP 8.2. （We don't need the 4.x branch now, right?）

harikt:
Cool. Thank you.

PR #216, opened 2026-07-24
Fix #157: do not treat AS inside an expression as a column alias

### LINKED ISSUE #157: Quoting problem with cast() on postgres
I'm on postgres and am writing a select that converts an integer field to varchar, so I can compare using "like".  The 2.x branch misquotes the cast structure's datatype.
```php
<?php
$select = $this->queryFactory->newSelect();
$select->cols(['street_number'])
        ->from('addresses')
        ->where('cast(street_number as varchar) like ?', '10%');
return $select->getStatement();
```
```sql
SELECT
    street_number
FROM
    "addresses"
WHERE
    cast(street_number AS "varchar) like :_1_"
```

harikt:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL quoting so `AS` inside `CAST(... AS ...)` expressions is no longer treated as a column alias separator.
  * Preserved correct quoting for real aliases after cast expressions.
  * Improved generated query output when using cast expressions in `WHERE` clauses (including `LIKE` with bound parameters).
* **Tests**
  * Added coverage for cast handling and alias parsing edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   on MySQL, which does not support DEFAULT VALUES), letting the database
   apply column defaults. Fixes #149.
 
+- [FIX] The quoter no longer treats an `AS` inside an expression (e.g.
+  `cast(col as varchar)`) as a column alias, which produced misquoted
+  SQL. Fixes #157.
+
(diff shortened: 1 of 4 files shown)


harikt:
@coderabbitai review

Issue #158, opened 2018-02-26
Idea: Parity across database models


designermonkey:
A discussion of an idea more than an issue.

I just ran across an issue where I didn't realise the `ON DUPLICATE KEY UPDATE` didn't work on SQLite.

As there are ways to achieve the same thing in SQLite, would it be a good idea to try for feature parity across models and implement the database specific way of achieving the same result?

harikt:
Hi there, 

Are you looking for the replace functionality ? I believe depending on the drivers additional features are already in it. If you think something is missing please do send a PR.

designermonkey:
I did not know about `REPLACE`, I will have a go with that.

harikt:
The specific surprise here — `ON DUPLICATE KEY UPDATE` not working on
SQLite — is addressed in `6.x` as of #236.

**SQLite and PostgreSQL now have the equivalent.** Both gained
`onConflict()` plus `doUpdateCol()`, `doUpdateCols()`, `doUpdate()` and
`doUpdateWhere()`, building `ON CONFLICT ... DO UPDATE`:

```php
$insert = $queryFactory->newInsert();
$insert
    ->into('users')
    ->cols(['email' => 'alice@example.com', 'name' => 'Alice'])
    ->onConflict('email')
    ->doUpdateCols(['name']);
```

```sql
INSERT INTO "users" (
    "email",
    "name"
) VALUES (
    :email,
    :name
)
ON CONFLICT ("email") DO UPDATE SET
    "name" = excluded."name"
```

**And where a dialect genuinely cannot do something, it now says so.**
That is the part most relevant to this issue. Previously, five of six
unsupported combinations died as a PHP fatal "call to undefined method",
which is exactly the experience you described. Those are now stubs on the
Common classes that throw
`Aura\SqlQuery\Exception\BadMethodCallException`, so an unsupported feature
is a catchable, named error rather than a crash or a silent difference.

**Full parity is not the goal, though**, and it is worth being explicit
about where it stops:

- **SQL Server is deliberately excluded from upsert.** Its equivalent is
  `MERGE` — a different statement shape, needing `HOLDLOCK` to avoid a
  race. Forcing it into the Insert path would be a worse API than not
  having it.
- **A portable `orReplace()` is not possible.** `REPLACE` (MySQL) and
  `INSERT OR REPLACE` (SQLite) are the same delete-then-insert operation
  and already share a method name. PostgreSQL and SQL Server have no
[...]

PR #219, opened 2026-07-25
Fix #172: unify insert-ignore as ignore() across dialects

### LINKED ISSUE #172: To add IGNORE statement to INSERT query library has different interfaces

harikt:
Fixes #172; related to #158.

Insert-ignore is now spelled `ignore()` on every dialect that supports it:

| Dialect | `ignore()` renders | Notes |
|---|---|---|
| MySQL | `INSERT IGNORE` | unchanged |
| SQLite | `INSERT OR IGNORE` | unchanged; `orIgnore()` now a deprecated alias |
| Pgsql | `ON CONFLICT DO NOTHING` | **new** — rendered after the values, before any `RETURNING` clause |
| Sqlsrv | throws `Exception\BadMethodCallException` | unchanged; T-SQL has no equivalent (now pinned by a test) |

`Sqlite\Update` also gains `ignore()` to match `Mysql\Update::ignore()`, with `orIgnore()` deprecated as an alias there too.

Semantic caveat, called out for reviewers: MySQL's `IGNORE` suppresses a range of errors (duplicate keys, bad enum values, etc.), while Postgres's `ON CONFLICT DO NOTHING` only skips conflict (duplicate-key) rows. They are equivalent for the common "insert if not exists" use case, which is what #172 asks for.

Tests: Pgsql `ignore()` asserted enabled, disabled, and combined with `RETURNING`; Sqlite `Update::ignore()` asserted alongside the legacy `orIgnore()`; Sqlsrv asserted to keep throwing. 461 tests, 828 assertions passing.

## Summary by CodeRabbit

- **New Features**
  - PostgreSQL INSERT queries now support `ignore()` to generate `ON CONFLICT DO NOTHING`.
  - SQLite UPDATE queries now support the standardized `ignore()` method.

- **Deprecations**
  - SQLite’s existing `orIgnore()` method remains available but is deprecated in favor of `ignore()`.

- **Compatibility**
  - SQL Server continues to report that ignore behavior is unsupported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,14 @@
   `cast(col as varchar)`) as a column alias, which produced misquoted
   SQL. Fixes #157.
 
+- [ADD] Insert-ignore is now spelled `ignore()` on every dialect that
+  supports it: Pgsql\Insert gains `ignore()`, rendering the Postgres
+  equivalent `ON CONFLICT DO NOTHING` (before any RETURNING clause), and
+  Sqlite\Update gains `ignore()` matching Mysql\Update; the Sqlite
+  `orIgnore()` methods remain as deprecated aliases. Sqlsrv, which has
+  no equivalent, keeps throwing Exception\BadMethodCallException.
+  Fixes #172; related to #158.
+
 - [CHG] An Update with no columns now throws
   Aura\SqlQuery\Exception\LogicException with
   a clear message, instead of a TypeError; an UPDATE with an empty SET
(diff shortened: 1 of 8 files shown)


harikt:
@coderabbitai review.

harikt:
@coderabbitai review

PR #220, opened 2026-07-25
Fix #177: allow # in quoted identifiers for DB2 / IBM i

### LINKED ISSUE #177: # in column name
Working with an exsisting DB structure and they have # special char all over their column names in a DB2 database. 

Using 
->orderBy( 'A.G01GL#' )

results in the sql statement being 

ORDER BY
   "A"."G01GL"# ASC 

Any advice on how to get around this?

harikt:
Fixes #177.

DB2 / IBM i allow `#` in identifiers, in any position (e.g. `A.G01GL#`). The quoter's `replaceNamesIn()` regex only matched `[a-z_][a-z0-9_]*`, so the match stopped before the `#` and produced broken SQL:

```
ORDER BY "A"."G01GL"# ASC   -- before
ORDER BY "A"."G01GL#" ASC   -- after
```

The identifier character class now includes `#` in any position, for **both** the table and column parts, and the `\b` anchors are replaced with `(?<![\w#])` / `(?![\w#])` lookarounds — `\b` cannot work adjacent to `#`, because `#` is a non-word character.

The regex suggested in the issue comments only extended the column part: with it, `lib#.tab` stays completely unquoted, and `#lib.tab` mis-quotes as `#"lib"."tab"` (the leading `\b` starts a match right after the `#`). Both cases are covered here, with tests pinning all five `#` positions: trailing/leading column, trailing/leading table, and mid-name.

Behavior is unchanged for identifiers without `#`; the existing suite passes as-is. One theoretical edge: a MySQL inline comment glued directly to a column with no space (`t.col#comment`) is now absorbed into the quoted name — previously it produced `"t"."col"#comment`, which was equally broken, so nothing working is affected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed quoting for fully qualified identifiers containing `#`.
  * Ensured identifiers such as `table.col#` are quoted correctly without breaking aliases or related name handling.

* **Tests**
  * Added coverage for identifiers containing `#` in multiple positions and database-specific quoting scenarios.

* **Documentation**
  * Added an entry describing the fix in the upcoming release notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@
   `cast(col as varchar)`) as a column alias, which produced misquoted
   SQL. Fixes #157.
 
+- [FIX] The quoter now recognizes `#` as an identifier character (legal
+  on DB2 / IBM i, in any position), so `table.col#` quotes as
+  `"table"."col#"` instead of the broken `"table"."col"#`. Fixes #177.
+
 - [CHG] An Update with no columns now throws
   Aura\SqlQuery\Exception\LogicException with
   a clear message, instead of a TypeError; an UPDATE with an empty SET
--- a/src/Common/Quoter.php
+++ b/src/Common/Quoter.php
@@ -284,19 +284,20 @@ class Quoter implements QuoterInterface
             return $text;
         }
 
-        $word = "[a-z_][a-z0-9_]*";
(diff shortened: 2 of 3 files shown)


harikt:
@coderabbitai review.

PR #231, opened 2026-07-26
Add Postgres LATERAL joins (port of #184)

harikt:
Ports @golgote's #184 to 6.x. His two commits are cherry-picked with original authorship intact; the follow-up commit adds tests, docs, and a correctness fix.

## The fix

The original emitted no `ON` clause when no condition was given, which PostgreSQL rejects. Verified against PostgreSQL 18.4:

| join type | no `ON` | `ON true` |
| --- | --- | --- |
| (plain) / INNER / LEFT | syntax error | ok |
| CROSS / NATURAL | ok | syntax error |

So `lateralJoinSubSelect()` now renders `ON true` when no condition is given, except for CROSS and NATURAL which forbid an `ON` clause.

## Usage

```php
$select
    ->cols(['dept.name', 'top.name AS top_earner'])
    ->from('dept')
    ->lateralJoinSubSelect(
        'left',
        'SELECT name FROM employee
         WHERE employee.dept_id = dept.id
         ORDER BY salary DESC LIMIT 1',
        'top'
    );
```

The signature matches `joinSubSelect()`.

## Tests

Unit tests in `tests/Pgsql/SelectTest.php` (condition, Select-object sub-select with bind values, no-condition, cross, duplicate-ref) plus three integration tests that execute the generated SQL against a real server — the point being that the bug was invalid SQL, which a string-comparison test alone would not have caught.

633 unit + 78 integration tests pass; PHPStan clean.

## Not included

MySQL 8.0.14+ also supports `JOIN LATERAL` (verified locally on 9.7.1), and SQL Server has `CROSS APPLY` / `OUTER APPLY` as its equivalent. Both left for a follow-up to keep this PR to what #184 proposed.

Closes #184.

## Summary by CodeRabbit

- **New Features**
  - Added PostgreSQL `JOIN LATERAL` support via `lateralJoinSubSelect()` for sub-selects (string or query), with correct aliasing, bind handling, and join-type behavior.
  - Renders `ON true` when no condition is provided for applicable lateral joins; omits `ON` for `CROSS`/`NATURAL`.
  - Rejects supplying a condition for `CROSS` or `NATURAL` lateral joins with a `LogicException`.
- **Documentation**
  - Updated `CHANGELOG` and PostgreSQL SELECT docs with usage and SQL examples.
- **Tests**
  - Added integration and unit coverage for SQL generation, execution, bindings, and error/edge cases (including duplicate references).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@
   `cast(col as varchar)`) as a column alias, which produced misquoted
   SQL. Fixes #157.
 
+- [ADD] Pgsql\Select gains `lateralJoinSubSelect()`, rendering
+  `JOIN LATERAL` against an aliased sub-select so the sub-select can
+  reference columns from the tables to its left. The signature matches
+  `joinSubSelect()`. PostgreSQL requires an `ON` clause on every LATERAL
+  join except CROSS and NATURAL, so when no condition is given for the
+  other join types, `ON true` is rendered. Thanks to @golgote for the
+  original implementation in #184.
+
 - [ADD] Insert-ignore is now spelled `ignore()` on every dialect that
   supports it: Mysql\Insert renders `INSERT IGNORE` and Sqlite\Insert
   renders `INSERT OR IGNORE` (both as before); Pgsql\Insert gains
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,8 +43,11 @@
   reference columns from the tables to its left. The signature matches
   `joinSubSelect()`. PostgreSQL requires an `ON` clause on every LATERAL
   join except CROSS and NATURAL, so when no condition is given for the
-  other join types, `ON true` is rendered. Thanks to @golgote for the
-  original implementation in #184.
+  other join types, `ON true` is rendered; conversely, passing a condition
+  to a CROSS or NATURAL join now throws
+  Aura\SqlQuery\Exception\LogicException, since PostgreSQL rejects an `ON`
+  clause there and the statement could only fail at execute time. Thanks
+  to @golgote for the original implementation in #184.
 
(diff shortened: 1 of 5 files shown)


harikt:
Good catch — fixed in db3626b.

You're right that `isUnconditionalJoin()` only suppressed the default and let an explicit condition through. I initially pushed back on the grounds that `join()` and `joinSubSelect()` have always behaved this way (`join('cross', 't2', 't1.c1 = t2.c1')` emits `CROSS JOIN "t2" ON …`, equally invalid), and that no single rule fits every dialect. The first point is real but weak — consistency with a decade-old permissive method isn't worth preserving in brand-new API with no callers. The second was just wrong: the correct framing is that no single *hard-coded* rule fits, which an overridable hook solves.

So `lateralJoinSubSelect()` now throws `Exception\LogicException` when a condition is given to a join type that rejects `ON`. The check runs before `addTableRef()`, so a rejected call leaves no table reference behind — covered by `testLateralJoinSubSelect_rejectedJoinLeavesNoTableRef()`.

The rule has to be dialect-aware. Verified against PostgreSQL 18.4 and MySQL 9.7.1:

| join type | PG `... ON` | MySQL `... ON` |
| --- | --- | --- |
| NATURAL | error | error |
| CROSS | error | **OK** |
| LEFT / INNER | ok | ok |

`NATURAL` is universal, but MySQL accepts `CROSS JOIN LATERAL … ON` since CROSS and INNER are synonyms there. Hence a separate `joinForbidsCondition()` alongside `isUnconditionalJoin()` — the trait carries the strict Postgres rule and `Mysql\Select` overrides it in the follow-up (#233).

Coverage added for both explicit-condition cases as you suggested, plus the no-table-ref-left-behind case.

harikt:
@coderabbitai fullreview

PR #232, opened 2026-07-26
Add MySQL LATERAL joins

harikt:
Adds `JOIN LATERAL` support to `Mysql\Select`, on top of the Postgres work in #231.

**Base is `6.x`, so this diff also contains #231's commits** (including @golgote's original two, with authorship preserved). Whichever of the two merges second will need a rebase.

## Change

`lateralJoinSubSelect()` lives in a new `Common\LateralJoinTrait`, used by both `Pgsql\Select` and `Mysql\Select`.

```php
$select
    ->cols(['dept.name', 'top.name AS top_earner'])
    ->from('dept')
    ->lateralJoinSubSelect(
        'left',
        'SELECT name FROM employee
         WHERE employee.dept_id = dept.id
         ORDER BY salary DESC LIMIT 1',
        'top'
    );
```

The signature matches `joinSubSelect()`.

## The dialects disagree, so one hook is overridable

Verified against real servers (PostgreSQL 18.4, MySQL 9.7.1) rather than from docs:

| join type | PG no `ON` | PG `ON` | MySQL no `ON` | MySQL `ON` |
| --- | --- | --- | --- | --- |
| (plain) / INNER | error | ok | ok | ok |
| LEFT | error | ok | error | ok |
| CROSS | ok | **error** | ok | **ok** |
| NATURAL | ok | error | n/a | error |

Two separate concerns, two methods:

- `isUnconditionalJoin()` — whether to supply the default `ON true`. CROSS/NATURAL on both dialects, so it stays in the trait. The original #184 code omitted `ON` entirely, which every dialect above rejects.
- `joinForbidsCondition()` — whether an explicit condition is rejected outright, throwing `Exception\LogicException` (raised by CodeRabbit on #231). NATURAL is universal, but **CROSS diverges**: MySQL accepts `CROSS JOIN LATERAL … ON` because CROSS and INNER are synonyms there, so `Mysql\Select` overrides it to reject NATURAL only.

The rejection check runs before `addTableRef()`, so a rejected call leaves no table reference behind.

## Caveats

`LATERAL` needs MySQL 8.0.14+, and **MariaDB has no `LATERAL` at all** despite sharing the `mysql` query objects. The integration tests skip on MariaDB and on MySQL < 8.0.14 via `PDO::ATTR_SERVER_VERSION`, and `docs/mysql.md` calls it out.

## Tests

Shared assertions live in `Common\LateralJoinTestTrait`, used by both dialects' `SelectTest` — they cannot go in `Common\SelectTest`, whose SQLite and Sqlsrv subclasses have no `LATERAL`. The CROSS-with-condition case is deliberately *not* shared: `Pgsql\SelectTest` asserts it throws, `Mysql\SelectTest` asserts it renders. Integration tests execute the generated SQL against real MySQL and Postgres, including the relaxed CROSS case — the point being that the original bug was invalid SQL, which string-comparison tests alone would not catch.

648 unit + 82 integration tests pass; PHPStan clean.

Closes #184.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@
   `cast(col as varchar)`) as a column alias, which produced misquoted
   SQL. Fixes #157.
 
+- [ADD] Pgsql\Select gains `lateralJoinSubSelect()`, rendering
+  `JOIN LATERAL` against an aliased sub-select so the sub-select can
+  reference columns from the tables to its left. The signature matches
+  `joinSubSelect()`. PostgreSQL requires an `ON` clause on every LATERAL
+  join except CROSS and NATURAL, so when no condition is given for the
+  other join types, `ON true` is rendered. Thanks to @golgote for the
+  original implementation in #184.
+
 - [ADD] Insert-ignore is now spelled `ignore()` on every dialect that
   supports it: Mysql\Insert renders `INSERT IGNORE` and Sqlite\Insert
   renders `INSERT OR IGNORE` (both as before); Pgsql\Insert gains
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,8 +43,11 @@
   reference columns from the tables to its left. The signature matches
   `joinSubSelect()`. PostgreSQL requires an `ON` clause on every LATERAL
   join except CROSS and NATURAL, so when no condition is given for the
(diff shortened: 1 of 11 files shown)


harikt:
@coderabbitai review

PR #233, opened 2026-07-26
Add MySQL LATERAL joins

harikt:
Follow-up to #231, which deliberately kept to Postgres. **Stacked on `port-lateral-joins`** — retarget to `6.x` once #231 merges, so the diff shown here stays MySQL-only.

## Change

`lateralJoinSubSelect()` moves out of `Pgsql\Select` into a new `Common\LateralJoinTrait`, used by both `Pgsql\Select` and `Mysql\Select`. No behavior change for Postgres.

## The dialects disagree, so one hook is overridable

Verified against real servers (PostgreSQL 18.4, MySQL 9.7.1):

| join type | PG no `ON` | PG `ON` | MySQL no `ON` | MySQL `ON` |
| --- | --- | --- | --- | --- |
| (plain) / INNER | error | ok | ok | ok |
| LEFT | error | ok | error | ok |
| CROSS | ok | **error** | ok | **ok** |
| NATURAL | ok | error | n/a | error |

Two separate concerns, so two methods:

- `isUnconditionalJoin()` — whether to supply the default `ON true`. CROSS/NATURAL on both dialects, so it stays in the trait.
- `joinForbidsCondition()` — whether an explicit condition is rejected (added in #231). NATURAL is universal, but **CROSS diverges**: MySQL accepts `CROSS JOIN LATERAL … ON` because CROSS and INNER are synonyms there. So `Mysql\Select` overrides it to reject NATURAL only.

## Caveats

`LATERAL` needs MySQL 8.0.14+, and **MariaDB has no `LATERAL` at all** despite sharing the `mysql` query objects. The integration tests skip on MariaDB and on MySQL < 8.0.14 via `PDO::ATTR_SERVER_VERSION`, and `docs/mysql.md` calls it out.

## Tests

Shared assertions live in `Common\LateralJoinTestTrait`, used by both dialects' `SelectTest` — they cannot go in `Common\SelectTest`, whose SQLite and Sqlsrv subclasses have no `LATERAL`. The CROSS-with-condition case is deliberately *not* shared: `Pgsql\SelectTest` asserts it throws, `Mysql\SelectTest` asserts it renders. Four MySQL integration tests execute the generated SQL against a real server, including the relaxed CROSS case, so the divergence is proven rather than asserted.

648 unit + 82 integration tests pass; PHPStan clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,16 +38,21 @@
   `cast(col as varchar)`) as a column alias, which produced misquoted
   SQL. Fixes #157.
 
-- [ADD] Pgsql\Select gains `lateralJoinSubSelect()`, rendering
-  `JOIN LATERAL` against an aliased sub-select so the sub-select can
-  reference columns from the tables to its left. The signature matches
-  `joinSubSelect()`. PostgreSQL requires an `ON` clause on every LATERAL
-  join except CROSS and NATURAL, so when no condition is given for the
-  other join types, `ON true` is rendered; conversely, passing a condition
-  to a CROSS or NATURAL join now throws
-  Aura\SqlQuery\Exception\LogicException, since PostgreSQL rejects an `ON`
-  clause there and the statement could only fail at execute time. Thanks
-  to @golgote for the original implementation in #184.
+- [ADD] Pgsql\Select and Mysql\Select gain `lateralJoinSubSelect()`,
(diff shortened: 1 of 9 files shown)


harikt:
Duplicate of #232, which targets 6.x directly. Closing in favour of that one.

Issue #124, opened 2017-01-24
Upsert


cdekok:
This is perhaps a bit db specific but it would be nice if there was an option for upsert in pgsql, seems there is something in mysql for it too not sure about other db's

pmjones:
I think #115 might cover that.

harikt:
@pmjones that looks only for mysql. May be pgsql needs to add something similar.

cdekok:
In postgres it's 
`ON CONFLICT DO UPDATE` or `ON CONFLICT DO NOTHING`
https://www.postgresql.org/docs/9.5/static/sql-insert.html

pmjones:
@cdekok Excellent -- can you post a comment on PR #115 to that effect? The contributor there might like to know.

cdekok:
@pmjones I just looked at the code but I think this is for different functionality, in mysql it would be 
`ON DUPLICATE` https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html
replace causes a delete before insert which could cause issues: http://stackoverflow.com/a/9168948

pavarnos:
ok. i wasn't aware of the implications (re the SO answer). lets abandon / delete #115 and start again with ON DUPLICATE KEY UPDATE. @cdekok do you want to make the new PR?

cdekok:
@pavarnos I could give it a shot, the one thing that would be hard is that with postgres you can specify the column or constraint on which the conflict occurs but on mysql it's always thrown on a primary key / unique index.
I don't really see how we could keep a consistent interface for them, any ideas?

pavarnos:
There is already something in the sqlite driver https://github.com/auraphp/Aura.SqlQuery/blob/3.x/src/Sqlite/Insert.php
https://www.sqlite.org/lang_conflict.html
#115 aimed to use the same method signature for MySQL and sqlite so unit tests could swap drivers with no compatibility layer needed. Would be cool if we could do the same for postgres. What about a method signature like this:
`function onDuplicateKey($sql = 'driver specific sql with sensible default')`
?
so the default is do nothing, then
```
Mysql\Insert::onDuplicateKey($sql = 'ON DUPLICATE KEY UPDATE');
Sqlite\Insert::onDuplicateKey($sql ='ON CONFLICT REPLACE');
Pgsql\Insert::onDuplicateKey($sql = 'ON CONFLICT DO UPDATE');
```

... just thinking out loud...

pmjones:
[...]

PR #236, opened 2026-07-26
Add ON CONFLICT upsert for Postgres and SQLite

harikt:
## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL and SQLite `INSERT ... ON CONFLICT` upsert support, including `DO NOTHING`, `DO UPDATE SET`, conflict targets (columns or named constraints where supported), and `DO UPDATE ... WHERE`.
  * Added fluent helpers for building `DO UPDATE SET` assignments.
* **Bug Fixes**
  * Unsupported `ignore()`, `orReplace()`, and upsert operations now consistently throw clear exceptions.
  * Added validation for illegal combinations and invalid/empty conflict targets (including SQLite `OR ...` flag restrictions).
* **Documentation**
  * Expanded PostgreSQL, SQLite, and MySQL docs; improved contributing guide and doc-example verification.
* **Tests**
  * Added PostgreSQL/SQLite integration coverage plus cross-dialect negative tests and doc-example checker config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,11 @@
   conflict target is Postgres-only -- SQLite accepts a column list and
   nothing else, and rejects that syntax -- so passing it to Sqlite\Insert
   throws Exception\BadMethodCallException instead of building SQL the
-  database cannot parse. Addresses the Postgres half of #124.
+  database cannot parse. An empty target -- `onConflict([])`,
+  `onConflict('')`, a blank column in the array, or the bare keyword
+  `ON CONSTRAINT` -- throws Exception\InvalidArgumentException, rather than
+  rendering `ON CONFLICT ()` for the database to reject. Addresses the
+  Postgres half of #124.
 
 - [CHG] Calling `ignore()`, `orReplace()` or the upsert methods on a
   dialect that does not support them now throws
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,8 +78,11 @@
   renders `ON CONFLICT (...) DO NOTHING` instead of `INSERT OR IGNORE`.
   Note that a raw `doUpdate()` expression must qualify any column it names
   on Postgres, which reads a bare name as ambiguous between the target
-  table and `excluded`; SQLite accepts either. Addresses the Postgres half
-  of #124.
+  table and `excluded`; SQLite accepts either. The `ON CONSTRAINT <name>`
+  conflict target is Postgres-only -- SQLite accepts a column list and
+  nothing else, and rejects that syntax -- so passing it to Sqlite\Insert
+  throws Exception\BadMethodCallException instead of building SQL the
+  database cannot parse. Addresses the Postgres half of #124.
 
 - [CHG] Calling `ignore()`, `orReplace()` or the upsert methods on a
   dialect that does not support them now throws
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,34 @@
   deprecated aliases. Sqlsrv, which has no equivalent, keeps throwing
   Exception\BadMethodCallException. Fixes #172; related to #158.
 
+- [ADD] Pgsql\Insert and Sqlite\Insert gain an upsert API:
+  `onConflict()` sets the conflict target (a column, an array of columns,
+  or `ON CONSTRAINT <name>`), and `doUpdateCol()`, `doUpdateCols()`,
+  `doUpdate()` and `doUpdateWhere()` build the `DO UPDATE SET` clause.
+  `doUpdateCols()` refers to the proposed row through the `excluded`
+  pseudo-table, `doUpdateCol()` binds a separate value, and `doUpdate()`
+  takes a raw expression. The two dialects share Common\OnConflictUpdate
+  Trait and Common\BuildOnConflictTrait, since SQLite adopted the Postgres
+  grammar in 3.24. Both databases require a conflict target for DO UPDATE,
+  and combining `ignore()` with the doUpdate methods is contradictory;
+  either throws Exception\LogicException at build time. On SQLite the
+  clause cannot be mixed with the `OR` flags, and `ignore()` with a target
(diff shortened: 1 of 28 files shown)


harikt:
For reference the plan that was used to make changes is below. 

# Proposals for Modernizing `INSERT OR UPDATE` & `INSERT OR IGNORE` in Aura.SqlQuery

This document outlines and compares two design paths for improving conditional insert/upsert operations in Aura.SqlQuery 6.x and beyond, updated with verified feedback regarding quoting, database-specific warnings, and concurrency correctness.

---

## Plan A: Driver-Specific Extensions (Non-Breaking / Evolutionary)

This plan retains the query builder's core philosophy of mapping closely to the underlying database dialect. It introduces the missing PostgreSQL and SQLite upsert features using database-specific syntax.

### API Design
- **MySQL**: Retains existing `onDuplicateKeyUpdateCol()`, `onDuplicateKeyUpdateCols()`, `onDuplicateKeyUpdate()`.
- **PostgreSQL & SQLite**: Introduce new methods via a shared `OnConflictUpdateTrait`:
  ```php
  $insert->into('users')
         ->cols(['email' => 'alice@example.com', 'name' => 'Alice'])
         ->onConflict('email')
         ->doUpdateCols(['name'])
         ->doUpdate('updated_at', 'NOW()');
  ```
- **SQL Server**: Unchanged (throws `BadMethodCallException` for ignore/replace/upsert).

### Quoting and the `EXCLUDED` Pseudo-Table
In PostgreSQL, referencing values proposed for insertion uses the `EXCLUDED.col` prefix. Passing `EXCLUDED.col` through `Quoter::quoteNamesIn()` converts it to `"EXCLUDED"."col"`, which fails with a `missing FROM-clause entry for table "EXCLUDED"` error because the double-quotes enforce case-sensitive matching on a real relation rather than the keyword. 

*Note: In SQLite, identifier quoting is case-insensitive (accepting `"EXCLUDED"."col"`), meaning this quoting constraint is imposed solely by PostgreSQL.*

Instead of altering the shared `Quoter.php` class—which is highly delicate and has repeatedly needed bug fixes in 6.0 (e.g., #157 for AS-in-expressions, #177 for # in identifiers, and @@session variables)—the builder itself can emit lowercase `excluded.<<col>>` directly when rendering the update columns from `doUpdateCols()`.
* **Example Builder Emission**:
  ```sql
  ON CONFLICT ("email") DO UPDATE SET "name" = excluded."name"
  ```
Since `excluded` in lowercase is accepted either quoted or bare on PostgreSQL (verified on PG 18.4; the `ON CONFLICT` and `excluded` pseudo-table features have been present since PG 9.5) and SQLite, this completely bypasses the quoter regex for this prefix and keeps the change entirely localized to the builder. Raw expressions via `doUpdate('col', 'expr')` remain the responsibility of the caller (retaining the same contract as MySQL's `onDuplicateKeyUpdate()`).

---

## Plan B: Unified Portable API (Breaking / Revolutionary)

This plan tries to abstract away database-specific dialects using a single high-level `upsert()` and portable `ignore()` API.

### API Design
```php
public function upsert(array $uniqueBy, array $updateCols = [], array $customUpdate = []);
```

### Critical Implementation Hurdles & Drawbacks

#### 1. MySQL/MariaDB Deprecation and Compatibility Split
Plan B's compiler would target `ON DUPLICATE KEY UPDATE col = VALUES(col)`. However:
- **MySQL 8.0.20+**: Deprecates the `VALUES()` function (Warning 1287), suggesting row aliases instead (`... AS new ON DUPLICATE KEY UPDATE col = new.col`).
- **MariaDB**: Does not support the row alias syntax and still requires `VALUES()`.
- **Conclusion**: Since the `'mysql'` driver in Aura.SqlQuery is shared by both engines, the compiler would have to query or guess the server type/version to avoid emitting deprecated or syntactically invalid SQL.

#### 2. SQL Server `MERGE` Concurrency and Correctness Issues
Implementing a portable upsert for SQL Server using `MERGE` without explicit locking is race-prone:
* Between the `WHEN NOT MATCHED` check and the subsequent insert, concurrent sessions can insert matching rows, causing duplicate-key violations or deadlocks under high load.
* A safe, concurrent SQL Server upsert requires explicit `HOLDLOCK` or `SERIALIZABLE` hints. Emitting a race-prone statement by default is a correctness regression.

#### 3. Leaky Portability in `$customUpdate`
If raw SQL is passed to the third parameter (e.g., `['updated_at' => 'NOW()']`), portability breaks immediately because `NOW()` does not exist in SQL Server (which requires `GETDATE()` or `SYSDATETIME()`).

#### 4. Interface and Subclass Breaking Changes
* Adding `upsert()` to `InsertInterface` introduces a breaking change to the interface surface, breaking all third-party implementers of the interface.
* `InsertInterface` does *not* declare `ignore()`. Instead, `ignore()` lives on `Common\Insert` and the dialect classes. To make `ignore()` portable on SQL Server, it would need to accept key columns. Changing its signature would break subclasses that override the method rather than the interface itself, but it remains a significant backward-compatibility concern.

---

## Recommendations & Roadmap

1. **Adopt Plan A for 6.0**:
   * Implement the PostgreSQL and SQLite conflict update syntax via `OnConflictUpdateTrait`.
   * Implement builder-side emission of `excluded.<<col>>` to avoid touching `Quoter.php`.
   * Keeps existing APIs intact, matches the library philosophy, and safely resolves the missing portion of issue #124.
2. **Move Stub Changes (Work Item 3)**:
   * Keep the stub changes on their own clean branch off `6.x` to isolate testing.
3. **Future Considerations for Plan B (6.1 / 7.x)**:
   * If a database-agnostic option is added in the future, it should be implemented as a separate query class (e.g. `Merge` or `Upsert`) generated via `QueryFactory` rather than overloading the `Insert` class. This avoids cluttering `InsertInterface` and isolates the concurrency/locking requirements of SQL Server's `MERGE`.

harikt:
@coderabbitai review

harikt:
@coderabbitai fullreview

PR #242, opened 2026-07-27
Keep the upsert binds through a bulk insert row

### LINKED ISSUE #241: Bulk-insert placeholders bypass bind-collision tracking
Follow-up to #238. The collision detection added there covers `cols()`,
conditions (per clause), sub-selects, closures and the upsert methods, but
bulk-insert placeholders are renamed and banked outside that tracking, so one
path still loses a value silently.

`finishCol()` moves each finished row's values into `$bind_values_bulk` under
`<name>_<row>`, and `getBindValues()` merges that array over `$bind_values`.
Neither step goes through `bindValueFrom()`, so nothing records that
`:status_0` is taken — and the merge is last, so it wins.

### Reproduce

```php
$insert = (new QueryFactory('pgsql'))->newInsert();
$insert
    ->into('t')
    ->cols(['status' => 'row0'])
    ->addRow(['status' => 'row1']);

$insert
    ->onConflict('id')
    ->doUpdateCol('status')
    ->doUpdateWhere('t.note = :status_0', ['status_0' => 'from the condition']);
```

```sql
INSERT INTO "t"
    ("status")
VALUES
    (:status_0),
    (:status_1)
ON CONFLICT ("id") DO UPDATE SET
    "status" = excluded."status"
WHERE
    "t"."note" = :status_0
```

```
getBindValues() === ['status_0' => 'row0', 'status_1' => 'row1']
```

The condition bound `:status_0` to `'from the condition'`, and that value is
gone. The `WHERE` now tests the first row's inserted value instead, with no
error raised — the statement is well-formed and runs.

Verified on the `6.x` line with #240 applied.

### Why it escapes the check

The condition's bind goes through `bindValueFrom('status_0', …, 'having'/'where')`
and finds the name free: at that moment `$bind_values` really does not contain
`status_0`, because the banked value lives in `$bind_values_bulk`. The clash
only materialises in `getBindValues()`, which merges the two arrays long after
any check could run.

### Rough shape of a fix

The bulk array needs to carry sources the way `$bind_values` does — a parallel
`$bind_sources_bulk` recorded in `finishCol()`, consulted by `bindValueFrom()`
and reconciled in `getBindValues()`. Worth checking at the same time whether a
user-supplied column named `status_0` alongside a bulk `status` column collides
by the same route.

### How likely is this?

Low. It needs a bulk insert whose auto-generated suffix name is then bound
explicitly by another clause, which mostly means a caller who knows the
`_<row>` convention and reuses it by accident. Filed for completeness rather
than urgency — the failure is silent, which is the part worth fixing.

harikt:
Part of #241 — the second symptom only, so please don't close #241 on merge.

`finishRow()` cleared `bind_values` wholesale, and `build()` calls it for the
last row, so a bulk insert dropped the upsert's own binds:

```php
$insert->into('t')->cols(['a' => 1])->addRow(['a' => 2])
       ->onDuplicateKeyUpdateCol('a', 3);
```

rendered `:a__on_duplicate_key` but bound only `a_0`/`a_1`, so `execute()`
failed with `HY093: Invalid parameter number`. Same for Pgsql
`doUpdateCol()` and `doUpdateWhere()` — bulk insert plus any upsert was broken.
Loud failure, not wrong data.

Fix: `finishCol()` returns the name it banked and `finishRow()` clears only
those, instead of emptying the array.

Unit tests for both dialects, plus integration tests running a colliding
two-row bulk upsert against live MySQL and PostgreSQL. Mutation-tested:
restoring the wholesale clear fails 2 unit tests and both integration tests;
never clearing the banked name fails 23. 742 unit, 107 integration, 17 doc
examples, PHPStan clean.

Still open in #241: the original report, bulk placeholders bypassing collision
tracking, which needs a parallel `$bind_sources_bulk`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed bulk insert statements so conflict-update values remain correctly bound across all inserted rows.
  - Improved handling of placeholder values during bulk inserts, preventing valid update parameters from being cleared.

- **Tests**
  - Added coverage for MySQL duplicate-key updates and PostgreSQL conflict updates.
  - Verified generated statements, retained parameters, conditional updates, and final database results for multi-row inserts.

### DIFF
--- a/src/Common/Insert.php
+++ b/src/Common/Insert.php
@@ -389,6 +389,13 @@ class Insert extends AbstractDmlQuery implements InsertInterface
      * Finishes off the current row in a bulk insert, collecting the bulk
      * values and resetting for the next row.
      *
+     * Only the names this row banked are cleared. Clearing the whole of
+     * bind_values would take the rest of the query's placeholders with it --
+     * the upsert values and their WHERE conditions, which belong to the
+     * statement as a whole rather than to any one row. build() finishes the
+     * last row, so those names would go missing from a statement that still
+     * spells them, and execute() would fail on the unbound placeholder.
+     *
      * @return null
      *
      */
@@ -399,12 +406,13 @@ class Insert extends AbstractDmlQuery implements InsertInterface
         }
 
         foreach ($this->col_order as $col) {
-            $this->finishCol($col);
+            $name = $this->finishCol($col);
+            if ($name !== null) {
+                unset($this->bind_values[$name], $this->bind_sources[$name]);
+            }
         }
 
         $this->col_values = array();
-        $this->bind_values = array();
-        $this->bind_sources = array();
     }
 
     /**
@@ -413,7 +421,8 @@ class Insert extends AbstractDmlQuery implements InsertInterface
      *
      * @param string $col The column to finish off.
      *
-     * @return null
+     * @return string|null The placeholder name this row banked, which the
+     * caller clears; null for a raw value, which never had one.
      *
      * @throws InvalidArgumentException on named column missing from row.
      *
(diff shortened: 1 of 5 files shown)


harikt:
@coderabbitai review .

PR #245, opened 2026-07-28
Track the placeholder names a bulk insert banks

### LINKED ISSUE #241: Bulk-insert placeholders bypass bind-collision tracking

harikt:
Fixes #241 — the half left open by #242.

A bulk insert renames each row's placeholders to `<name>_<row>` and banks them outside `bind_values`, so the collision check never saw them, and the merge in `getBindValues()` came last and won:

```php
$insert->into('t')->cols(['status' => 'row0'])->addRow(['status' => 'row1'])
       ->onConflict('id')->doUpdateCol('status')
       ->doUpdateWhere('t.note = :status_0', ['status_0' => 'from the condition']);
// bind values: {"status_0":"row0","status_1":"row1"} — the condition's value gone,
// the WHERE silently testing row 0's data instead
```

Fix: a parallel `$bind_sources_bulk`, checked in both directions — a clause binding a banked name, and a row banking over a name a clause holds. Columns of the row being built are exempt: `a_1` alongside `a` is renamed out of the way before the merge, so it never collides. Hand binds now correctly overwrite a banked value, which they previously could not.

Four guards, each mutation-tested to failure. One of them survived the first pass, which turned out to be a missing test rather than dead code — a live column spelled the same as an earlier row's banked name, reachable by calling `getBindValues()` before the statement is built — now pinned.

Integration test on Postgres, plus docs and a changelog entry. 747 unit, 108 integration, 17 doc examples, PHPStan clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Bulk-insert generated placeholder names are now included in collision detection.
  - Conflicts involving banked bulk-insert placeholders now raise a `LogicException` with a clearer collision source.
  - Bulk placeholder name handling is corrected for ordering, resets, and `bindValue()` interactions.
- **Documentation**
  - Added a “Bulk Insert Rows” section explaining per-row placeholder renaming, reservation behavior, and workarounds for collisions.
- **Tests**
  - Added collision and PostgreSQL integration coverage for bulk-insert placeholder conflicts and `ON CONFLICT ... DO UPDATE` conditions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,19 @@
   `DB_MYSQL_DSN` / `DB_PGSQL_DSN` / `DB_SQLSRV_DSN` are set; see
   CONTRIBUTING.md.
 
+- [BRK] The placeholder names a bulk insert generates are now tracked like
+  any others. Each row's placeholders are renamed `<name>_<row>`, and those
+  banked names were invisible to the collision check while still winning the
+  merge in getBindValues() -- so after
+  `cols(['status' => ...])->addRow([...])`, a condition binding `:status_0`
+  had its value silently replaced by row 0's, and the statement ran against
+  the wrong one. Either order now throws
+  Aura\SqlQuery\Exception\LogicException; bind the condition under a name no
+  row can generate. Columns of the row being built are exempt, since a column
+  named `a_1` alongside `a` is renamed out of the way before the merge.
+  Binding by hand with bindValue() is unaffected and now correctly overwrites
+  a row's value, where before the banked value won regardless. Fixes #241.
+
 - [FIX] A bulk insert combined with an upsert no longer loses the upsert's
   bound values. Finishing a row cleared every bound value, not just that
   row's, and building the statement finishes the last row -- so
--- a/docs/other.md
+++ b/docs/other.md
@@ -94,6 +94,34 @@ reserved, so do not bind them yourself. A column literally named
 `onDuplicateKeyUpdateCol('name', ...)`, and `name__on_conflict` collides with
 `doUpdateCol('name', ...)`. Both throw the same `LogicException`.
 
+### Bulk Insert Rows
+
+A bulk insert renames each row's placeholders, appending the row number: two
+rows of `cols(['status' => ...])` bind `:status_0` and `:status_1`. Those
+generated names are reserved in the same way, so binding one from another part
+of the query throws:
+
+```php
+$insert = $queryFactory->newInsert();
+
+try {
+    $insert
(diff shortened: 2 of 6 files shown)


harikt:
@coderabbitai fullreview.

harikt:
@coderabbitai fullreview

Issue #183, opened 2021-03-24
Where with parentheses


designermonkey:
Maybe I'm missing something... How would I replicate the following?

```
where (
    find_in_set(tests.`compound.group`, :compound_groups) or
    find_in_set(tests.`compound.name`, :compound_names)
)
and (
    find_in_set(isolates.`species.genus`, :species_genera) or
    find_in_set(isolates.`species.name`, :species_names)
)
```

The parentheses are important in the above, and I can't figure out how I should do it with SQLQuery.

harikt:
Both halves of this now work on the `6.x` branch, and the builder produces the exact SQL you asked for.

Pass a closure to `where()` to get a parenthesized group:

```php
$select->where(function ($w) {
    $w->where('find_in_set(tests.`compound.group`, :compound_groups)')
      ->orWhere('find_in_set(tests.`compound.name`, :compound_names)');
});

$select->where(function ($w) {
    $w->where('find_in_set(isolates.`species.genus`, :species_genera)')
      ->orWhere('find_in_set(isolates.`species.name`, :species_names)');
});
```

Output:

```sql
SELECT
    *
FROM
    `tests`
WHERE
    (
        find_in_set(tests.`compound.group`, :compound_groups)
        OR find_in_set(tests.`compound.name`, :compound_names)
    )
    AND (
        find_in_set(isolates.`species.genus`, :species_genera)
        OR find_in_set(isolates.`species.name`, :species_names)
    )
```

Two things were involved:

1. **Grouping.** Closure grouping already emitted the parentheses; it just wasn't documented, which is presumably why it wasn't findable. The docs now cover it (#223).

2. **Hand-quoted dotted names.** This was the part that was genuinely broken. The quoter treated `` `compound.group` `` as a qualified name and re-quoted it to `` `compound`.`group` ``, turning one column whose name contains a dot into a table/column pair. It now leaves an already-quoted identifier exactly as written, on MySQL and SQL Server (#225). Unit and integration tests cover it.

Closing as fixed. Thanks for the report, and sorry it sat this long.

PR #246, opened 2026-07-28
Read a comma-separated table list as a list

### LINKED ISSUE #160: Cannot use FROM with multiple table names
The following is expected:

```sql
UPDATE "themes", "entities"
SET
    "themes"."name" = :themes.name
WHERE
    "entities"."uuid" = :uuid
```

instead when doing `->table(Tables::THEMES . ', ' . Tables::ENTITIES)` I get:

```sql
UPDATE "themes," "entities"
SET
    "themes"."name" = :themes.name
WHERE
    "entities"."uuid" = :uuid
```

and when doing `->table(Tables::THEMES . ' , ' . Tables::ENTITIES)` I get:

```sql
UPDATE "themes" "," "entities"
SET
    "themes"."name" = :themes.name
WHERE
    "entities"."uuid" = :uuid
```

harikt:
Fixes #160.

`from('t1, t2')` was read as a name and its alias and quoted whole, giving `"t1," "t2"` — an identifier no database has, failing at execute time with nothing to say why. Same on all four dialects, and for `Update::table()` and `Delete::from()`.

- **SELECT** builds the list: each table quoted and reference-checked on its own, identical to calling `from()` once per table.
- **UPDATE/DELETE** throw `LogicException` naming the sub-select alternative. There is no portable statement to build — MySQL `UPDATE a, b`, PG/SQLite `UPDATE a … FROM b`, SQL Server `UPDATE a … FROM a JOIN b`, and `DELETE FROM a, b` is valid nowhere.

The split is quote-aware, so a comma inside an identifier stays part of the name. Making that reachable meant also letting `quoteName()` leave an already-quoted identifier alone — the #183 rule, which until now applied only inside expressions, so `table('"odd,name"')` rendered `""odd,name""`. It is dialect-aware: MySQL still wraps a double-quoted name.

Lives in a new `Common\\TableListTrait` used by Select/Update/Delete. Nothing public changed — `QuoterInterface` is untouched, so third-party implementers are unaffected.

Five guards, each mutation-tested to failure. Integration test runs the multi-table FROM against MySQL, PostgreSQL and SQLite (SQL Server in CI). 787 unit, 115 integration, 19 doc examples (up from 17 — the two new ones carry `sql` blocks so the checker actually verifies them), PHPStan clean.

## Summary by CodeRabbit

* **New Features**
  * `SELECT` `from()` now supports comma-separated table names (including aliases) in a single call, while preserving quoted identifiers.

* **Bug Fixes**
  * Fixed identifier handling for comma-separated inputs so already-quoted names aren’t re-processed and generated SQL stays valid.
  * `UPDATE` (`table()`) and `DELETE` (`from()`) now strictly require a single table and throw a clear error suggesting a sub-select alternative.

* **Documentation**
  * Updated `FROM`, `UPDATE`, and `DELETE` docs with the new multi-table behavior and examples.

* **Tests**
  * Added coverage for splitting, quoted-comma behavior, duplicate detection, and the new validation errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,21 @@
   conditions. Part of #241; the remaining half of that issue, bulk
   placeholders bypassing collision tracking, is still open.
 
+- [FIX] Naming several tables in one string no longer produces an identifier
+  no database has. `from('t1, t2')` was read as a name and its alias and
+  quoted whole, giving `"t1," "t2"`; a Select now builds the list, quoting
+  and reference-checking each table, so it is the same as calling `from()`
+  once per table. An identifier you quoted yourself is left as you wrote it,
+  as it already was inside expressions, so a comma within it stays part of
+  the name.
+
+  Update::table() and Delete::from() take a single table and now throw
+  Aura\SqlQuery\Exception\LogicException for a list, naming the sub-select
+  alternative. There is no portable statement to build: MySQL writes a
+  multi-table update as `UPDATE a, b SET ...`, PostgreSQL and SQLite as
+  `UPDATE a SET ... FROM b`, and SQL Server as `UPDATE a SET ... FROM a JOIN
+  b`, while `DELETE FROM a, b` is valid nowhere. Fixes #160.
+
 - [FIX] An Insert with no columns no longer throws a TypeError; it now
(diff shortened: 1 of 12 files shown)


harikt:
@coderabbitai fullreview

PR #247, opened 2026-07-28
Fix 189 union query

harikt:
## Summary by CodeRabbit

* **New Features**
  * `union()` and `unionAll()` now accept a complete `Select` query as the next UNION branch.
  * Union branch SQL and bound values are captured at add time (later changes to the provided query won’t affect the already-added branch).
  * Placeholder/bind-name collisions are validated, and `resetUnions()` clears prior union state.

* **Documentation**
  * Updated UNION docs with chaining rules, binding behavior, and error conditions after a query-based union tail.

* **Tests**
  * Added unit and integration coverage for query-based unions, bindings, subselect usage, resets, and invalid usage.

PR #249, opened 2026-07-28
Hold placeholders from every union branch

### LINKED ISSUE #248: Union of three or more branches loses earlier branches' placeholder claims
A union of three or more branches silently overwrites a placeholder that an earlier branch still binds.

```php
$select->cols(['c1'])->from('t1')->where('a = :a', ['a' => 1])
    ->union()->cols(['c2'])->from('t2')->where('b = :b', ['b' => 2])
    ->union()->cols(['c3'])->from('t3')
    ->where('a = :a', ['a' => 999]);
```

The third branch takes `:a` without complaint and `getBindValues()` returns `['a' => 999, 'b' => 2]`, while the first branch's retained SQL still reads `a = :a`. That branch then executes against 999 instead of 1 — the wrong-results case #238 set out to catch.

`Select::resetAfterRendering()` rebuilds `bind_sources` from nothing on each `union()`, scanning only `end($this->union)`, so every name spelled by an earlier branch is handed back as free. Two branches are unaffected (the one retained branch is also the newest), which is why the #238 tests did not cover it; `:b` above, from the branch that was scanned, still throws correctly.

Affects 6.x, and predates the `union($select)` work in #247.

harikt:
Fixes #248.

**Bug.** From the third union branch on, a placeholder an earlier branch still binds can be taken by a later one without a word:

```php
$select->cols(['c1'])->from('t1')->where('a = :a', ['a' => 1])
    ->union()->cols(['c2'])->from('t2')->where('b = :b', ['b' => 2])
    ->union()->cols(['c3'])->from('t3')
    ->where('a = :a', ['a' => 999]);
```

Branch 1's retained SQL still reads `a = :a`, so it executes against 999.

**Fix.** `resetAfterRendering()` rebuilds `bind_sources` from nothing on each `union()` and was reading only `end($this->union)`, handing back every name the earlier branches spell. It now scans all retained branches.

**Coverage.** Two tests, both running across all five dialects: the collision above, and a guard that all three branches' values survive. The collision test fails 5x before the fix; restoring `end($this->union)` fails it again. Suite, PHPStan, doc examples and integration are green.

Trial-merges clean with #247 (999 tests green together); the two can land in either order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed UNION placeholder “hold/claim” tracking across three or more branches to prevent silent overwrites.
  * Placeholder detection now avoids placeholder-like text inside SQL string literals and comments, reducing false collision errors.
  * UNION placeholder collisions and reuse now raise clearer errors, and quoted-identifiers are handled so placeholders inside them aren’t scanned.
  * Improved dialect-specific placeholder parsing (MySQL, PostgreSQL, and SQL Server).

* **Tests**
  * Expanded UNION regression coverage for held-placeholders, collision scenarios, and correct bind merging across branches.
  * Added broader parameterized fixtures covering quoting, literals/comments, and edge cases (including unclosed quotes behavior).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,7 +92,7 @@
   above exists to prevent, arriving one branch later. The claims are
   rebuilt from the retained SQL on each `union()`, and only the newest
   branch was being read. Two-branch unions were unaffected. As elsewhere,
-  branches may share a name so long as they share its value.
+  branches may share a name so long as they share its value. Fixes #248.
 
 - [FIX] Naming several tables in one string no longer produces an identifier
   no database has. `from('t1, t2')` was read as a name and its alias and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,14 @@
   branch was being read. Two-branch unions were unaffected. As elsewhere,
   branches may share a name so long as they share its value. Fixes #248.
 
+- [FIX] A union branch no longer holds a placeholder name that only its
+  string literals or comments spell. `where("name = ':a'")` binds nothing
+  by that name, and holding it reported a collision against the next
+  branch's legitimate `:a`. Literals and comments are blanked before the
+  names are read; anything unterminated is read as SQL instead, since
+  keeping a name costs at worst a collision report where losing one lets a
+  later branch overwrite the SQL silently.
+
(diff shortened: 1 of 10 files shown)


harikt:
@coderabbitai fullreview

Issue #130, opened 2017-03-06
CTEs


mathroc:
Hi,

now that mysql will support WITH queries too, is there plan to add support for Common Table Expressions ?

maybe something along those lines :

```php
$select = $query_factory->newSelect();

$cte = $query_factory->newSelect();
$cte->cols('zim')->from('table2');

$select->with('cte1', $cte);

$select->cols(array('foo', 'bar AS barbar'))
       ->from('table1')
       ->from('cte1')
       ->where('cte1.zim = 99');
```

pmjones:
This is an excellent idea. It may mess up [Atlas.Orm](https://github.com/atlasphp/Atlas.Orm), which uses `with()` to eager-load relationships -- but since Atlas is alpha, it can change to use `eager()` or something like that instead of the WITH keyword.

@mathroc send along a PR as soon as you can, for the databases that support WITH, and I'll review!

harikt:
Done in #251 (6.x): `with()` and `withRecursive()` on Select, taking a Select or a raw string, with an optional column list. Your example builds as written.

CTEs on INSERT, UPDATE and DELETE are the follow-on, tracked in #261 and implemented in #262.

PR #251, opened 2026-07-29
Add a WITH clause to SELECT (#130)

harikt:
Adds common table expressions to SELECT, requested in #130.

```php
$select->with('well_paid', $earners)
       ->cols(['name'])
       ->from('well_paid');
```

- `with($name, $spec, $cols = [])` and `withRecursive()`; `$spec` is a Select or a raw string, rendered on the spot the way a branch passed to `union()` is.
- The clause prefixes the statement, so a union writes it once above every branch and each branch may name the CTEs.
- SQL Server infers the recursion and rejects the keyword, so `withRecursive()` writes a plain `WITH` there. The same PHP is valid on every dialect.
- `resetWith()` releases the clause and the names it claimed.

The placeholder bookkeeping is the fiddly part. A CTE belongs to the statement rather than to a branch, so its claims have to survive the reset `union()` performs — a name only the CTE holds passes to it, and one a rendered branch also spells is held by both, so either reset hands the name over instead of freeing it. Freeing it would let a later clause rebind a placeholder the retained SQL still spells, silently.

A branch passed to `union()` may not bring a `WITH` of its own; it would render as `UNION WITH ... SELECT`, so it is reported.

Covered by unit tests across all four dialects, collision tests for the bind ownership, and integration tests including a recursive CTE that counts 1→3 against a real server. Every guard was mutation-tested — deleted, with the named test confirmed to fail.

## Summary by CodeRabbit

* **New Features**
  * Added support for Common Table Expressions (CTEs) in SELECT queries through `with()` and `withRecursive()`.
  * Supports multiple CTEs, column lists, raw SQL definitions, recursive queries, unions, and bound values.
  * Added `resetWith()` to clear CTE definitions.
  * Added SQL Server-specific handling for recursive CTE syntax and query limits.

* **Bug Fixes**
  * Improved placeholder collision handling across CTEs, unions, and query resets.

* **Documentation**
  * Added comprehensive CTE usage and SQL Server guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 ## 6.0.0 (unreleased)
 
+- [ADD] SELECT queries take common table expressions, via with() and
+  withRecursive():
+
+        $select->with('well_paid', $earners)
+               ->cols(['name'])
+               ->from('well_paid');
+
+  The CTE is given as a Select or as a raw string, with an optional column
+  list, and a Select is rendered on the spot the way a branch passed to
+  union() is; the values it bound come with it. RECURSIVE belongs to the
+  clause rather than to one CTE, so a statement mixing recursive and plain
+  members spells it once. SQL Server infers the recursion and rejects the
+  keyword, so withRecursive() renders a plain WITH there and the same PHP is
+  valid on every dialect.
+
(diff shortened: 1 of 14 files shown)


harikt:
@coderabbitai review.

PR #255, opened 2026-07-30
Use short array syntax in tests

harikt:
## Summary by CodeRabbit

* **Refactor**
  * Modernized test code to use PHP short array syntax consistently.
  * Updated query-building inputs, expected bind values, and test data providers across supported database drivers.
  * Preserved existing SQL expectations, assertions, and test behavior.

PR #256, opened 2026-07-30
Ignore the formatting-only sweeps in git blame

harikt:
The two `array()` sweeps (#254, #255) rewrote 782 lines without changing
behaviour, so `git blame` now credits them for lines whose author and reason
are years older. `.git-blame-ignore-revs` lists both commits and blame looks
through them.

Before and after, on a line the tests sweep touched:

```
off: dfdc144a  Hari K T         2026-07-30   $expect = [
on:  5cad658b  Rasmus Schultz   2014-01-28   $expect = [
```

GitHub reads the file by itself. Local blame needs `git config
blame.ignoreRevsFile .git-blame-ignore-revs` once per clone, which CONTRIBUTING
now documents along with how to add future entries.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for excluding formatting-only commits from `git blame` results.
  * Documented how to configure Git and add formatting commit revisions for clearer change history.
  * Added references to formatting conventions in source code and tests.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,8 @@
+# Commits that changed formatting only, and so should not be blamed for the
+# lines they touched. See CONTRIBUTING.md for how to make git use this file.
+
+# Use short array syntax in src
+8966e329e35bbff966ebc387264b796e6bfca8e4
+
+# Use short array syntax in tests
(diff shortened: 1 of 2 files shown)


harikt:
@coderabbitai review

PR #257, opened 2026-07-30
Correct wrong docblocks and raise PHPStan to level 3

harikt:
First step of the 6.x typing work. No code changes and no behaviour change --
docblocks only, plus the PHPStan level.

Three kinds of tag were wrong, not merely missing:

* 12 `@return null` on methods that return nothing. `null` means "returns the
  value null", which is why PHPStan reported 16 `return.missing`; the tag is
  `@return void`.
* 6 properties documented with `@param`, or with an empty `@var`. These were
  unparseable, so every one of those properties had no type at all.
* `$quoter` documented as the concrete `Common\Quoter` in two places while the
  constructor already accepts `Common\QuoterInterface`. Nothing in src/ calls a
  method outside the interface, so the interface is the honest type.

`$builder` is redeclared with a narrowed `@var` in the 10 classes that call
their own dialect's builder methods. A `@var` rather than a native property
type on purpose: typed properties are invariant, so once the parent declares
one, a subclass cannot redeclare it at all -- narrower or untyped -- and this
narrowing becomes impossible.

That takes PHPStan from level 1 to level 3 clean. Level 4 reports 8 more
docblocks that claim a narrower type than the code accepts (a `@param string`
on a method testing `instanceof Closure`, three comparing against null); the
code is right and the tags are wrong, so that is the next PR.

Checks unchanged: 1303 unit, 132 integration, 24 doc examples. Each half was
mutation-tested by reverting it and confirming PHPStan fails again.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal type declarations and documentation for query builders and properties.
  * Clarified method return types and interface annotations across query components.
  * Improved support for static analysis by raising the configured analysis level.
* **Bug Fixes**
  * Replaced implicit builder properties with explicit declarations to improve reliability and compatibility.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,7 +1,9 @@
-# Start at the highest level that passes clean; raise the level as the
-# 6.x native-types work lands (level 2+ currently fails on $this->builder
-# being typed as AbstractBuilder while queries call dialect-builder methods).
+# Start at the highest level that passes clean; raise the level as the 6.x
+# typing work lands. Level 4 currently reports 8 docblocks that claim a
+# narrower type than the code accepts -- a @param string on a method that
+# tests `instanceof Closure`, and three that compare against null -- so the
+# code is right and the tags are wrong. Fixing those is the next step.
 parameters:
-    level: 1
+    level: 3
     paths:
         - src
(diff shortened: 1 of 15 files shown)


harikt:
@coderabbitai review

PR #262, opened 2026-08-15
Take common table expressions on INSERT, UPDATE and DELETE

### LINKED ISSUE #261: CTEs on INSERT, UPDATE and DELETE
#130 asked for common table expressions, and its example is a SELECT. 6.x
implements `with()` / `withRecursive()` on Select only, so this is the
follow-on: every dialect this package targets also allows a WITH clause on
data-modifying statements.

Verified here against SQLite 3.53.2, in memory, no server needed:

```
WITH c AS (SELECT id, n FROM src) INSERT INTO t (id, n) SELECT id, n FROM c   -- accepted
WITH c AS (SELECT id FROM src WHERE id = 1) UPDATE t SET n = 'z' WHERE ...    -- accepted
WITH c AS (SELECT id FROM src WHERE id = 2) DELETE FROM t WHERE ...           -- accepted
```

Postgres, MySQL 8.0+ and SQL Server document support as well, though I have
not run those locally.

### What is already in place

`WithInterface` was deliberately kept separate from `SelectInterface` so that
the DML queries could implement it, and `WithTrait` holds the whole
implementation -- name collision handling, `resetWith()`, the recursive flag.
None of that is Select-specific.

### What is missing

Two pieces, both small and both mechanical:

1. `WithTrait` declares `abstract protected function subSelect()`, and the
   only implementation is `Common\Select::subSelect()`. That method needs
   just `bindValuesFromSelect()` plus indenting, and `bindValuesFromSelect()`
   already lives on `AbstractQuery` -- so `subSelect()` could move there and
   serve all four query types rather than being reimplemented three times.

2. `buildWith()` lives on `Common\SelectBuilder` and is called from
   `Select::build()`. The Insert, Update and Delete builders would need it,
   so it would move to `Common\AbstractBuilder`.

harikt:
Follow-on to the SELECT CTE work: `with()` and `withRecursive()` now work on
the data-modifying queries too. Fixes #261.

`WithTrait` moves to `AbstractDmlQuery`, `subSelect()` to `AbstractQuery`, and
`buildWith()` to `Common\AbstractBuilder`. New Sqlsrv DML builders so SQL
Server renders a plain `WITH` on all four query types.

MySQL takes no `WITH` on INSERT — it allows a CTE only inside the SELECT an
`INSERT ... SELECT` draws from, which this package does not build — so
`Mysql\Insert::with()` throws instead of emitting unexecutable SQL.

[BRK] `InsertInterface`, `UpdateInterface` and `DeleteInterface` extend
`WithInterface`, as `SelectInterface` already did.

Coverage: shared `WithTestTrait` on the three Common DML test classes, so every
dialect inherits it, plus integration tests that execute against a real server.
CI green on SQLite, MySQL 8.0/8.4, Postgres 15/17 and SQL Server 2019/2022.

## Summary by CodeRabbit

* **New Features**
  * Added common table expression (CTE) support for `INSERT`, `UPDATE`, and `DELETE` queries.
  * Added standard and recursive CTEs, including multiple CTEs, column lists, and parameter binding.
  * Added SQL Server support with dialect-appropriate CTE syntax.

* **Bug Fixes**
  * MySQL `INSERT` queries now clearly reject unsupported CTE usage.

* **Documentation**
  * Added usage examples and dialect-specific guidance for CTE-enabled data modification queries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@
   outside it now has four more methods to declare; extending the concrete
   query, or using Common\WithTrait, supplies them.
 
+- [BRK] AbstractQuery::getStatement() is abstract. Select and AbstractDmlQuery
+  both write clauses above the ones build() renders -- the union branches, the
+  WITH clause -- so nothing was left for the base implementation to do, and a
+  query type extending AbstractQuery directly now says how its statement is
+  assembled rather than inheriting `return $this->build();`.
+
 - [ADD] SELECT queries take common table expressions, via with() and
   withRecursive():
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,33 @@
 
 ## 6.0.0 (unreleased)
 
+- [ADD] INSERT, UPDATE and DELETE queries take common table expressions too,
+  via the same with() and withRecursive():
+
+        $delete->with('seniors', $seniors)
+               ->from('employee')
+               ->where('id IN (SELECT id FROM seniors)');
+
+  The clause is the one SELECT already had -- WithTrait moved up to
+  AbstractDmlQuery and buildWith() to Common\AbstractBuilder -- so a CTE
+  behaves the same wherever it is written: the sub-select is rendered on the
+  spot with the values it bound, RECURSIVE is spelled once for the clause,
+  and resetWith() releases the names.
+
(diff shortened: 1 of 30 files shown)


harikt:
@coderabbitai review

harikt:
@coderabbitai review

PR #263, opened 2026-08-22
Say what AbstractQuery's arrays hold, and raise PHPStan to level 5

harikt:
Docblock-only. Adds array shapes to AbstractQuery's 7 array properties and 8
method params/returns, and gives `inlineArray()` the docblock it never had.

With the file clean, PHPStan goes to level 5. That is what makes the inline
`@phpstan-ignore` on the `array_filter(..., 'strlen')` line in Select legal --
below level 5 the annotation is itself an `ignore.unmatchedIdentifier` error,
so the level and the annotation have to arrive together. Level 6 across `src`
drops from 161 errors to 145.

The bind arrays are keyed `int|string`, not `string`: `bindValues()` avoids
`array_merge()` precisely so integer keys survive, and they do --
`bindValues([1 => 'a', 'b' => 'c'])` keeps `1` as an integer key. The three
bind properties, and everything downstream that mirrors them, are shaped that
way. `$bind_sources` and `$bind_shared` hold `string` and never null, both
writes being guarded. `$flags` is `array<string, true>`; only `true` is ever
written.

No behaviour changes. Unit, integration, doc examples and analyse all green.

## Summary by CodeRabbit

- **Quality Improvements**
  - Raised the configured static analysis level to improve code quality checks.
  - Clarified type documentation for query bindings, clauses, flags, parameters, and return values.
  - Added documentation for inline array handling.
  - Documented a static-analysis exception in query rendering code.
  - No user-facing runtime behavior or public API changes were introduced.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -6,6 +6,6 @@
 # messages PHPStan emits for them are identical wherever they occur, so there
 # is nothing to key a pattern on.
 parameters:
-    level: 4
+    level: 5
     paths:
         - src
--- a/src/AbstractQuery.php
+++ b/src/AbstractQuery.php
@@ -415,7 +415,7 @@ abstract class AbstractQuery
      *
      * Reports two parts of the query claiming one placeholder name.
      *
-     * @param string $name The placeholder name.
+     * @param int|string $name The placeholder name.
      *
      * @param string $prior The part of the query holding the name.
      *
--- a/src/AbstractQuery.php
+++ b/src/AbstractQuery.php
@@ -220,7 +220,7 @@ abstract class AbstractQuery
      *
      * Binds a single value to the query.
      *
-     * @param string $name The placeholder name or number.
+     * @param int|string $name The placeholder name or number.
      *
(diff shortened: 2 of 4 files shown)


harikt:
Verified and fixed in ca32141.

Reproduced first: `bindValue(1, 'a')` is reported at level 5 for a consumer, and it reproduces at the branch base e10660c too, so the contract was already wrong before this PR rather than made wrong by it. But this is the PR that establishes `array<int|string, mixed>` for the bind arrays, and leaving `@param string $name` beside it in the same file is a contradiction, so it belongs here.

`bindValue()` and `bindValueFrom()` in AbstractQuery and `bindValue()` in QueryInterface now take `int|string`. The prose always said "the placeholder name or number" -- only the type disagreed.

The one on QueryInterface is what actually fixes the consumer, since calls resolve through the interface: reverting it alone brings the error back, while reverting AbstractQuery's two does not. Both are corrected for consistency.

harikt:
@coderabbitai review

harikt:
@coderabbitai review

PR #264, opened 2026-08-22
Say what the clause arrays hold, from setter through to builder

harikt:
Stacked on #263 -- review that one first.

Docblock-only. Shapes the WHERE, WITH, ORDER BY, table-list, LATERAL and
Pgsql RETURNING seams. Each is taken end to end -- interface, trait or
property, and the builder method that consumes it -- because a shape given
on only one side of a seam reads as a disagreement rather than as a missing
type.

The `$bind` params carry `array<int|string, mixed>` down from #263, which is
where that key type is established. `$with` is `array<string, string>`,
`$returning` and `$where` are `list<string>`, `$flags` is
`array<string, true>`, and the public setters that callers hand arbitrary
arrays to take `array<array-key, string>`.

Level 6 across `src` drops from 145 errors to 109. Unlike the property
shapes in #263, these are checked: retyping `buildWhere()`'s param to
`list<int>` reports 3 errors at level 5.

No behaviour changes. Unit, integration, doc examples and analyse all green.

## Summary by CodeRabbit

- **Documentation**
  - Clarified PHPDoc annotations for query-building parameters and return values.
  - Documented accepted array key and value types for filtering, bindings, ordering, common table expressions, joins, and column lists.
  - Added more precise type information for PostgreSQL `RETURNING` values.
  - Improved documentation consistency across query-building interfaces and traits.
  - No runtime behavior or public method signatures were changed.

### DIFF
--- a/src/Common/AbstractBuilder.php
+++ b/src/Common/AbstractBuilder.php
@@ -22,7 +22,7 @@ abstract class AbstractBuilder
      *
      * Builds the flags as a space-separated string.
      *
-     * @param array $flags The flags to build.
+     * @param array<string, true> $flags The flags to build.
      *
      * @return string
      *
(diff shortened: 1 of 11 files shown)


harikt:
@coderabbitai review

